PR #1, opened 2020-12-29
Implementation with Twind [fixed]

lukejacksonn:
In a conversation on twitter you remarked that you had given Twind a go and that it weren't behaving as you'd expected, which is something we couldn't stand to see (and probably meant we weren't documenting things properly) 😅 

So I have taken the time to fixup and rebase your branch, as a learning exercise for myself more than anything!

The result is now like for like with what you have in production, at least as far as I can tell. The changes I had to make were actually pretty minimal but it has certainly hinted at where are docs are lacking so thank you. 

Things you had already done perfectly:

- The dependencies `styled-components` and `twin.macro` had been replaced with `twind`
- The dev dependencies `autoprefixer`, `postcss` and `tailwind` had been removed
- No need to configure `babelMacros` with twin preset pointing at styled-components
- Setup next to perform SSR with Twind as documented in the Twind [next-example](https://github.com/tw-in-js/example-next)

Slight changes I had to make:

- Global styles (for the html/body and scrollbars) were moved into the `setup` function and extended the [`preflight`](https://github.com/tw-in-js/twind/blob/main/docs/setup.md#preflight)
- All instances of `tw="..."` were converted to ```className={tw`...`}``` (although we might support `tw` prop soon)
- The custom background image style was converted from a `style` prop into an [inline plugin](https://github.com/tw-in-js/twind/blob/main/docs/plugins.md#inline-plugins)
- Uses the `tw([...])` syntax when joining multiple rule sets, previously a comma was being inserted preventing styles being applied (you can read more about the Twind function signature [here](https://github.com/tw-in-js/twind/blob/main/docs/tw.md#function-signature))

I'm really happy that this was all that was "broken" and actually really pleased with the result. Again, thanks for this opportunity and do let us know if you have any further questions or feedback 🙏

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,8 +1,0 @@
-module.exports = {
-  presets: [['next/babel', { 'preset-react': { runtime: 'automatic' } }]],
-  plugins: [
-    'babel-plugin-macros',
-    '@babel/plugin-transform-react-jsx',
-    ['styled-components', { ssr: true }],
-  ],
-}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,23 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es2021": true,
+        "node": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:react/recommended"
+    ],
+    "parserOptions": {
+        "ecmaFeatures": {
+            "jsx": true
+        },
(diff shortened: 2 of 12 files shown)


lukejacksonn:
Oh shucks.. I see a flash of unstyled content and the global styles not applying on the deployed version (but its not showing locally) 😓 it is late here for me but I will check this out first thing tomorrow!

PR #3, opened 2021-12-13
feat: render to a stream to allow flushing headers

jacob-ebey:


tannerlinsley:
This is really interesting. I don't plan on merging, as you said, it doesn't do much for perf right now. But I can't wait to see suspense and streaming really come to life in the future.

PR #27, opened 2022-08-04
Update announcing-tanstack-query-v4.md

GreatState:
Fix 'network mode' link


tannerlinsley:
I’d like to keep it relative

PR #52, opened 2022-11-21
Don't play timer spinning animation for reduced motion users

zadeviggers:


crutchcorn:
Thanks so much for this change! A11Y is a huge priority and something we'll be putting a huge effort into soon (we first need to cleanup the site codebase and such so efforts aren't duplicated/lost).

crutchcorn (approved):

PR #55, opened 2022-12-07
feat: framework switcher

DamianOsipiuk:
Should be merged along https://github.com/TanStack/query/pull/4608

Co-authored-by: Aryan Deora <adeora@iu.edu>


app/components/FrameworkSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { Form, useMatches } from '@remix-run/react'
+import { useNavigate } from '@remix-run/react'
+
+// TODO: we could potentially pass available frameworks trough config
+const frameworks = [
+  { label: 'React', value: 'react' },
+  { label: 'Solid', value: 'solid' },
+  { label: 'Vue', value: 'vue' },
+  { label: 'Svelte', value: 'svelte' },
+  { label: 'Angular', value: 'angular' },
```
TkDodo, line 20: we won't really have angular in tanstack/query because it's in a community adapter:

https://tanstack.com/query/v4/docs/community/angular-query

is there a way to make this customizable ? Maybe we could bind it to the structure that is listed under `adapters` ?

Screenshot 2022-12-08 at 07 49 53

app/components/FrameworkSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+  angular: angularLogo,
+}
+
+const properCase = (str: string) => str[0].toUpperCase() + str.slice(1)
+
+export function FrameworkSelect() {
+  const navigate = useNavigate()
+  const matches = useMatches()
+  
+  const match = matches.find(m => m.id === 'routes/query/v4/docs/$framework')
```
TkDodo, line 37: given that we are in `app/components/FrameworkSelect.tsx` here, is it correct to have a hardcoded match upon query v4 here?

TkDodo (commented):
great work 🙌

app/components/FrameworkSelect.tsx
DamianOsipiuk: Framework is passed from upper scope via prop now.

app/components/FrameworkSelect.tsx
DamianOsipiuk: Available options in the switcher are now inherited based on sections from the docs config

TkDodo (approved):

tannerlinsley:
Are we good to merge here?

DamianOsipiuk:
@tannerlinsley Yup, but it has to go together with https://github.com/TanStack/query/pull/4608

tannerlinsley:
Are redirects being handled anywhere here?

Tanner Linsley

PR #59, opened 2022-12-26
fix: change replacement token

DamianOsipiuk:
Changing token used to perform content replacement due to prettier.
When formatting document token gets changed and would be broken, therefore it's better to align it to the format prefered by prettier


DamianOsipiuk:
@TkDodo

tannerlinsley:
Looks good. Ready to merge?

PR #60, opened 2022-12-29
✨ feat(Search): add data-theme attr to document

Moshyfawn:
Based on the [@docsearch/css](https://docsearch.algolia.com/docs/styling/) package[ _variables.css](https://github.com/algolia/docsearch/blob/2c64362067376a76575b7b5131972f71c920a406/packages/docsearch-css/src/_variables.css#L51).

A quick way to add dark mode to the DocSearch component.

Can obviously be improved, but I'd suggest to leave it until there's some work done on allowing to toggle between the themes as a user. @TkDodo needs his dark mode docs search when he's steaming!

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -47,8 +47,8 @@ export let links: LinksFunction = () => {
       media: '(prefers-color-scheme: dark)',
     },
     {
-			rel: "stylesheet",
-			href: docSearchStyles
+      rel: "stylesheet",
+      href: docSearchStyles
     },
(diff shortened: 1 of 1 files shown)


Moshyfawn:
Can also be done by adding [the stylesheet link](https://cdn.jsdelivr.net/npm/@docsearch/css@3) instead

TkDodo:
looks good. This is also what I do on my blog. Thank you ❤️

PR #62, opened 2023-01-04
feat(query): multi-version docs support

DamianOsipiuk:
- changed all hard-coded `v4` references to `version` route param
- fixed framework switcher redirects to try to target same page (ex. `react/installation` -> `vue/installation`)
- moved configurable `repo`, `defaultBranch`, `defaultVersion` to main `query` route file

@TkDodo This enables us to have both `v4` and `v5` docs available


TkDodo:
will we also have a `/latest/` in the url, similar to what `react-router` has?
- https://reactrouter.com/en/main

I'm just wondering how we'll handle external links that now point to `/v4` after we've released `v5`. permalinks are good because maybe a page only exists for v4 and not v5, but e.g. from my blog, I'd mostly want to link to the "latest" docs version without having to update links manually 🤔

DamianOsipiuk:
I think we could easily add `latest` as version and just map it to the proper branch.
In this case though i think it would be better to keep the `latest` in the url instead of redirecting.
Also when we switch to v5 we could add a banner similar to the current one for v3. Saying that the `latest version is now x click here to redirect to latest docs`

DamianOsipiuk:
@TkDodo 
- added handling for `latest`
- `latest` is now the `default` redirect from https://tanstack.com
- `latest` still shows as `v4` in the name on main `query` page
- `v4`, `v5` and other `v` should be handled automatically as long as there is a corresponding branch in `query` repo
- for now `v4` points to `main` branch, which is easily switchable to next version if needed

TkDodo (approved):

Issue #82, opened 2023-05-03
Sidebar covers only 80% of the screen on mobile


skirianov:
## Issue description
The Sidebar covers only 80% of the screen on mobile and suffers from double scroll (when sidebar content scroll is over, underneath content is being scrolled)

## Attachements
![Screenshot 2023-05-03 at 17 14 07](https://user-images.githubusercontent.com/74229951/235960845-bf4fb171-51a7-4b6b-bb5c-b022ba7a543b.png)
You can see at the bottom main content is visible

## Fix
I would like to be assigned to this issue and fix it

skirianov:
Opened a PR with fixes: #83

tannerlinsley:
Outdated.

PR #83, opened 2023-05-03
docs: fixed sidebar issue

skirianov:
PR for Issue #82  
Fixes: 
- Sidebar is now full height and is displaying properly on all devices in Chrome DevTools
- Body overflow is turned off when sidebar is displayed to prevent double scrollbar issues

**Why not simply remove the overflow from the body and keep scrolling only inside scroll containers?** 
In order to preserve the functionality of `<ScrollRestoration />` on mobile, I needed to keep body overflow visible. Otherwise, we would lost the "scroll back to top" functionality on navigation between pages

**Why do we set `document.body.style.overflow = 'visible'` inside `Docs.tsx`?**
Since the `toggleBodyOverflow` function works only when sidebar is toggled on/off and does not return the body overflow value to `visible` if navigated to the main page (clicking on the logo), we have to set it explicitly to `visible` in the component itself. 

☝️ if you are aware of a better approach, please let me know, I would be happy to implement it :)

### DIFF
--- a/app/components/Docs.tsx
+++ b/app/components/Docs.tsx
@@ -50,7 +50,6 @@ export function Docs({
   v2?: boolean
 }) {
   const matches = useMatches()
-  console.log(matches)
   const lastMatch = last(matches)
 
   const detailsRef = React.useRef<HTMLElement>(null!)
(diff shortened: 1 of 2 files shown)


skirianov:
@tannerlinsley can you take a look please?

PR #86, opened 2023-06-23
fix: use `<article>` for semantic

pacexy:


crutchcorn:
I'm unfortunately going to close this issue temporarily. :( 

I know we have immense work to put into A11Y/SEO for this project, but right now we need to prioritize consolidating code implementations; as the codebase is sorta all over the place right now.

Once we have our code structure better organized, it'll make it much easier to do code audits/reviews on A11Y and more. It's a huge priority for me so rest assured it will be done

PR #91, opened 2023-07-26
fix: place carbon ads on the right hand side and not left menu sidebar

okandas:
Before:

image

After:

image

Small fix,  place carbon ads on the right sidebar so that they do not interfere with the left sidebar menu navigation ..

### DIFF
--- a/app/components/Docs.tsx
+++ b/app/components/Docs.tsx
@@ -218,9 +218,6 @@ export function Docs({
       <div className="flex-1 flex flex-col gap-4 px-4 whitespace-nowrap overflow-y-auto text-base pb-[300px]">
         {menuItems}
       </div>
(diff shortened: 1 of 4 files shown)


tannerlinsley:
As long as we're doing this, let's move the ad image to the right and swap the image rounding from the top right to the top left corner.

okandas:
> As long as we're doing this, let's move the ad image to the right and swap the image rounding from the top right to the top left corner.

image

Done, not sure if these ads appear anywhere else so its tested to make sure it doesn't affect anything else ?

SeanCassiere:
This has been completed in https://github.com/TanStack/tanstack.com/commit/ac79e7b672f14df95fc2714c4a0d2afc9b670014.

PR #110, opened 2023-11-01
chore: Switch from yarn to pnpm

lachlancollins:
This project recommends setup with yarn, while the other TanStack projects use pnpm. A `pnpm-lock.yaml` file already existed in the repo, so this PR deletes (and ignores) `yarn.lock` and sets the `packageManager` field to pnpm.


lachlancollins:
@tannerlinsley could you please review this? I'm hoping it doesn't need any changes to the Vercel install command.

PR #113, opened 2023-11-05
Add missing framework subroutes

lachlancollins:
Form and Stores projects have their framework docs nested under an additional folder - I missed this when upgrading to the remix v2 routing structure in #111


lachlancollins:
@tannerlinsley

PR #131, opened 2023-12-31
Update the README.md with info about editing the docs locally

fulopkovacs:
Lately several people have asked about editing the `TanStack/form` docs and previewing them locally in the `#form` channel on Discord, so we updated `CONTRIBUTING.md` in that repo to address this: https://github.com/TanStack/form/pull/537.

@crutchcorn suggested that rather than doing this in every project, we could add the instructions to this one, and then reference it everywhere else.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,55 @@ pnpm dev
 ```
 
 This starts your app in development mode, rebuilding assets on file changes.
+
+## Editing and previewing the docs of TanStack projects locally
+
+The documentations for all TanStack projects except for `React Charts` are hosted on [https://tanstack.com](https://tanstack.com), powered by this Remix app.
+In production, the markdown doc pages are fetched from the GitHub repos of the projects, but in development they are read from the local file system.
+
+Follow these steps if you want to edit the doc pages of a project (in these steps we'll assume it's [`TanStack/form`](https://github.com/tanstack/form)) and preview them locally :
+
+1. Create a new directory called `tanstack`.
+
+```sh
+mkdir tanstack
+```
+
+2. Enter the directory and clone this repo and the repo of the project there.
+
+```sh
+cd tanstack
+git clone git@github.com:TanStack/tanstack.com.git
+git clone git@github.com:TanStack/form.git
+```
+
+> [!NOTE]
+> Your `tanstack` directory should look like this:
+>
+> ```
+> tanstack/
+>    |
+>    +-- form/
+>    |
+>    +-- tanstack.com/
+> ```
+
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,3 +18,55 @@ pnpm dev
...
+pnpm dev
+```
+
+4. Now you can visit http://localhost:3000/form/latest/docs/overview in the browser and see the changes you make in `tanstack/form/docs`.
+
+> [!NOTE]
+> The updated pages need to be manually reloaded in the browser.
+
+> [!WARNING]
+> You will need to update the `docs/config.json` file (in the project's repo) if you add a new doc page!
```
fulopkovacs, line 72: I've also seen people getting confused about updating `<project>/docs/config.json` (they don't know when to update it). If you want, I can add a few more lines about it (based on the source code of the Remix app).

crutchcorn: Let's do it 😄

fulopkovacs: So I looked into the usage of `config.json` files across the TanStack projects, but it seems like they are a bit different, so I think they should be documented in their own repos. 

Some examples of the differences:

| project | | file |
| --- |--- | --- |
| `Form` | has `menu` for core docs, `frameworkMenus` for the framework-specific docs | [config.json](https://github.com/TanStack/table/blob/89ea3efb82180b414fb19e203654d8257067150a/docs/config.json#L7) |
| `Query` | only has `menu` and framework-specific docs, no shared core docs | [config.json](https://github.com/TanStack/query/blob/e6cb6df5682d11d06e2421ae6a5281fe32f4de36/docs/config.json#L7) |
| `Table` | only has `menu` and _no_ framework-specific docs | [config.json](https://github.com/TanStack/table/blob/89ea3efb82180b414fb19e203654d8257067150a/docs/config.json#L7) |

tannerlinsley: I think we should have a single format that supports multi framework.

fulopkovacs: Sounds great! So far the `Form` config seems to be the one that could fit all the TanStack projects, but tomorrow I can go through all of them again to make sure this format covers everything. If it does, I'm happy to make a PR for the new, cross-project, multi-framework config format.

tannerlinsley: Excellent!

fulopkovacs: Ok, I checked all the projects, and it looks like the `Form` config format fits all of them if we add a `users` key to it (for TanStack Query). I opened up a PR draft with more info and a list of tasks. Can you check it, @tannerlinsley ? If you give me the green light, I'll start working on the tasks.

lachlancollins (approved):

Issue #129, opened 2023-12-23
The TanStack Store React quick start page has too many redirects


richardtknight:
### Describe the bug

The link to "[quick start](https://tanstack.com/store/latest/docs/framework/react/quick-start)" doesn't load anything, and if you open it in a new tab, eventually it resolves into an ERR_TOO_MANY_REDIRECTS error being thrown as per the screenshot attached.

### Your Example Website or App

https://tanstack.com/store/latest/docs/framework/react/quick-start

### Steps to Reproduce the Bug or Issue

1. Go to the TanStack store documentation [overview page](https://tanstack.com/store/latest/docs/overview)
2. Click "Quick Start" in the left hand menu
3. Wait...

### Expected behavior

As a user I expected the quick start page to load.

### Screenshots or Videos

<img width="1510" alt="Screenshot 2023-12-23 at 16 20 25" src="https://github.com/TanStack/store/assets/16280330/85c02f71-4e56-4cf0-9d0e-1c20f4415e1e">

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 91.1]

fulopkovacs:
https://github.com/TanStack/tanstack.com/pull/133 will fix this ([See this comment](https://github.com/TanStack/tanstack.com/pull/133/files/09bd3f53bd4c8ceb894d9b0a09137f9889a5488a#r1439069701)).


Basically, the example routes are not set up correctly. When the user loads `/store/latest/framework/react/quick-start` it matches the `store.$version.docs.framework.$framework.$examples._index.tsx` route, where `$examples` is a dynamic segment. 

> [!NOTE] 
> These are the value of the [route params](https://remix.run/docs/en/main/guides/data-loading#route-params): `{version: "latest", framework: "react", examples: "quick-start`}`

The problem is that this is the loader of `store.$version.docs.framework.$framework.$examples._index.tsx`:
```tsx
export const loader = (context: LoaderFunctionArgs) => {
  throw redirect(context.request.url.replace('/examples/', `/examples/simple`))
}
```
Since `context.request.url` does not contain `'/examples/'`, it'll just redirect the route to itself. This is where the infinite redirects come from.


The solution is simply to drop the `$` from `$examples` in examples' routes:
- `app/routes/store.$version.docs.framework.$framework.$examples._index.tsx` -> `app/routes/store.$version.docs.framework.$framework.examples._index.tsx`
[...]

PR #133, opened 2024-01-01
Fix the framework picker for form

fulopkovacs:
Two issues has been fixed for the framework picker of the Form project:
1. It was not working on core pages, like Overview, which is the first page the users see when they open the docs (I thought that the framework picker is not working at all because of this for a long time).
2. If the user selected a framework and then viewed a core doc page, it went back to React, regardless of the previously selected framework.

**:x: `main`**

https://github.com/TanStack/tanstack.com/assets/43729152/41af42a2-e8c7-48ac-a82c-6228c3630637

**:white_check_mark: This PR**

https://github.com/TanStack/tanstack.com/assets/43729152/bc389f45-9908-4e59-8c70-0ae0571ebe00


fulopkovacs:
@crutchcorn There are a couple of ideas, that that we could add to this PR:

1. Make these changes for `Store` too (at first glance, the structure of `Form` and `Store` look very similar)

2. Drop the `framework` string from the routes of the docs pages:
   - `"/docs/latest/framework/react/quick-start"` -> `"/docs/latest/react/quick-start"`
   - this would make the docs more consistent (the `Query` doc routes already look like this)
   - I think it's technically feasible, because we know all the names of the frameworks from the `config.json` file in the project's repo (I haven't thought deeply about this problem yet though)
  
  Ofc these are just suggestions, changing routes are scary, plus I know much less about the projects than you do. 🤣 Wdyt?

crutchcorn:
@fulopkovacs let's do #1 for sure. I'm less sure about #2 - @tannerlinsley what do you think?

app/routes/store.$version.docs.framework.$framework.examples._index.tsx
fulopkovacs, line 1: I think `examples` wasn't supposed to be a dynamic segment in this route, because `store.$version.docs.framework.$framework.$examples._index.tsx` matches `/store/latest/docs/framework/react/quick-start` (`$examples`: `"quick-start"`) and all the other framework-related doc pages.

app/routes/form.$version.docs.framework.$framework.$.tsx
```diff
@@ -1,20 +1,57 @@
...
+import type { GithubDocsConfig } from '~/routes/form'
 import { repo, getBranch } from '~/routes/form'
 import { DefaultErrorBoundary } from '~/components/DefaultErrorBoundary'
 import { seo } from '~/utils/seo'
 import { useLoaderData, useParams } from '@remix-run/react'
 import { Doc } from '~/components/Doc'
 import { loadDocs } from '~/utils/docs'
+import { fetchRepoFile } from '~/utils/documents.server'
 
 export const loader = async (context: LoaderFunctionArgs) => {
```
fulopkovacs, line 11: The body of the `loader` function is basically identical in this file and [`app/routes/store.$version.docs.framework.$framework.$.tsx`](https://github.com/TanStack/tanstack.com/pull/133/files#diff-cec2d983f6a6adc45d30d2ffe186f539471e0d67e4948e2c3e2b8b5ed15276caR11-R55). If we want to follow the DRY practices, we can extract the shared logic to a helper function.

crutchcorn (changes requested):
@fulopkovacs I noticed that we've seemingly fixed this by making the `core` docs duplicate through every framework.

```
core/quick-start.md -> routes/react/quick-start.md, routes/vue/quick-start.md, etc
```

This is a clever solution, but unfortunately one we can't move forward with. That would break our doc's SEO pretty bad, as there wouldn't be a canonical URL for `core` docs anymore

Instead, let's use some other kind of tool (maybe localstorage + a sidebar refactor?) to solve this issue.

fulopkovacs:
> @fulopkovacs I noticed that we've seemingly fixed this by making the `core` docs duplicate through every framework.
> 
> ```
> core/quick-start.md -> routes/react/quick-start.md, routes/vue/quick-start.md, etc
> ```
>
> This is a clever solution, but unfortunately one we can't move forward with. That would break our doc's SEO pretty bad, as there wouldn't be a canonical URL for `core` docs anymore
> 
> Instead, let's use some other kind of tool (maybe localstorage + a sidebar refactor?) to solve this 

Sure, I can explore other options. We can use `URLSearchParams` if we want to keep using SSR, or go down the client-first `localStorage` route, I'll check which one requires less changes!

crutchcorn:
You're awesome, thank you for your help and receptiveness to feedback. Keep up the amazing work! 🔥

fulopkovacs:
Ok, I went with `localStorage` for cleaner URLs. It's pretty late here, so this implementation might have mistakes, idk... I'll review it tomorrow and request a review from you when I'm ready!

app/routes/form.$version.docs.$.tsx
```diff
@@ -17,7 +17,7 @@ export const loader = async (context: LoaderFunctionArgs) => {
   })
 }
 
-export const meta: MetaFunction = ({ data }) => {
+export const meta: MetaFunction<typeof loader> = ({ data }) => {
```
fulopkovacs, line 20: I fixed the type errors around the `meta` functions in the files I worked on. Most of the remaining errors of this kind are getting fixed in https://github.com/TanStack/tanstack.com/pull/135.

app/routes/store.$version.docs.framework.$framework.examples._index.tsx
fulopkovacs, line 1: This fixes https://github.com/TanStack/tanstack.com/issues/129#issuecomment-1890062946

crutchcorn: This is awesome, thank you!

app/routes/store.tsx
```diff
@@ -118,7 +118,7 @@ export const useReactStoreDocsConfig = () => {
   const params = useParams()
   const version = params.version!
   const framework =
-    params.framework || localStorage.getItem('framework') || 'react'
+    params.framework || localStorage.getItem(storeLocalStorageKey) || 'react'
```
crutchcorn, line 121: Just checking on this again - could we keep it as `framework` rather than `framework-store` and `framework-form`?

crutchcorn (changes requested):

app/routes/form.tsx
```diff
@@ -118,8 +118,9 @@ export const useReactFormDocsConfig = () => {
   const match = matches[matches.length - 1]
   const params = useParams()
   const version = params.version!
+
   const framework =
-    params.framework || localStorage.getItem('framework') || 'react'
+    params.framework || localStorage.getItem(formLocalStorageKey) || 'react'
```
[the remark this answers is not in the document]
fulopkovacs: Yeah, no specific reason. I just assumed it's not a feature but a bug! I'm changing it back to `framework`.

crutchcorn (approved):
This is a great fix, good work!

PR #144, opened 2024-01-20
refactor: Move shared project config out of routes directory

lachlancollins:
Before, the routes would import constants from each other, such as `import { v8branch } from '~/routes/table.v8'`

This PR refactors any shared project config out of routes and into a `projects` directory.

This will make migrating to a different router easier, since the shared code is outside the routes directory, and can therefore be used without changes.


tannerlinsley:
Approved, but needs rebase

lachlancollins:
> Approved, but needs rebase

I've resolved the conflicts 👍

PR #148, opened 2024-01-22
refactor: Clean-up docs logic and layouts

lachlancollins:
- Fixed redirects for form/store base framework path (e.g. `/form/latest/docs/framework/react` would 500 error)
- Renamed some variables for clarity (e.g. `kind` -> `framework`)
- Removed some empty layouts (returning only `<Outlet />`)
- Replaced object spreading with JSX values on `<Docs />` components
- Fixed wrong project names for stackblitz iframes
- Added `$schema` as permitted schema property (very meta)
- Renamed `<Docs />` to `<DocsLayout />`
- Changed form/store docs redirect from `/<project>/latest` to `/<project>/latest/docs/overview`, actually reports an error when `config.json` is not found rather than silently redirecting


fulopkovacs (approved):
lgtm! 🚀 ☺️

PR #149, opened 2024-01-23
feat: Migrate table docs to v2 layout

lachlancollins:
- Uses layout and logic common to form and store
- Adds framework and version selector
- Adds docs which appear across all frameworks ("core" docs)
- Refactored `use<Project>DocsConfig` hook into shared `useDocsConfig`

See also: https://github.com/TanStack/table/pull/5295


app/routes/table.$.tsx
```diff
@@ -4,7 +4,7 @@ import { redirect } from '@remix-run/node'
 export const loader = (context: LoaderFunctionArgs) => {
   handleRedirects(context)
 
-  return redirect('/table/v8')
+  return redirect('/table/latest', 301)
```
fulopkovacs, line 7: Are we sure about having a permanent redirect here? I think some browsers cache 301 redirects indefinitely, so if a user visits a `/table/*` route, that is not matched by other routes today then they might be forever (I mean indefinitely 😄 ) redirected to `/table/latest` even after we add the page they were looking for.

This would happen to `/table/latest/something`:
```
❯ curl -I  http://localhost:3000/table/latest/something
HTTP/1.1 301 Moved Permanently
location: /table/latest
Date: Wed, 24 Jan 2024 20:15:14 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

I'd go with `302`, just to be safe.

app/routes/table.$version._index.tsx
```diff
@@ -44,7 +46,7 @@ const menu = [
         <FaBook className="text-lg" /> Docs
       </div>
     ),
-    to: './docs/guide/introduction',
+    to: './docs/introduction',
```
fulopkovacs, line 49: Are we moving this doc to a new place in the `tanstack/table` repo too? Right now we only have `./docs/guide/introduction`: https://github.com/TanStack/table/blob/main/docs/guide/introduction.md 

`/table/v8/docs/introduction` shows a `File does not exist` error message for me:
image

fulopkovacs (commented):
I left two comments, but otherwise it's super cool to see the codebase getting cleaned up!

app/routes/table.$.tsx
lachlancollins: Makes sense, I just copied what the other projects had in this file

app/routes/table.$version._index.tsx
lachlancollins: Yes sorry it's this PR here: https://github.com/TanStack/table/pull/5295

fulopkovacs: Ahhh, got it!

app/routes/table.$.tsx
fulopkovacs: Hmmm, I guess this is not a new problem then... 😆 Should we do something about it, or is it a lost cause already?

lachlancollins: Done!

fulopkovacs (approved):
LGTM 🚀🚀

PR #150, opened 2024-01-23
fix: Framework text colour with v2 docs layout

lachlancollins:
Currently, for the v2 layout (form/store), the react and "core" docs have text coloured, while other frameworks have their background coloured. I think this was a bug, so I've changed it so text is coloured - here is the difference.

## Before:
image

## After:
image


fulopkovacs (approved):
haha, I thought these were intentional, but I did find the taste of the designer rather peculiar...  🤣

PR #151, opened 2024-01-25
feat: Migrate query docs to v2 layout

lachlancollins:

See also: https://github.com/TanStack/query/pull/6763 (main), https://github.com/TanStack/query/pull/6764 (v4), https://github.com/TanStack/query/pull/6765 (v3)


app/projects/query.tsx
```diff
@@ -1,37 +1,83 @@
...
-const latestBranch = 'main'
+export type Framework = keyof typeof frameworks
 
-export const latestVersion = 'v5'
+export const createLogo = (version?: string) => (
+  <>
+    <Link to="/" className="font-light">
+      TanStack
+    </Link>
+    <Link to=".." className={`font-bold`}>
```
fulopkovacs, line 35: Just to be consistent with code 3 lines above:
```suggestion
    <Link to=".." className="font-bold">
```

fulopkovacs (changes requested):
One tiny comment, but otherwise it's good to go! No need to request a second review (unless GitHub requires it). ☺️

app/projects/query.tsx
lachlancollins: Fixed - it does need another review, sorry!

I'll do this for the other PRs too

fulopkovacs (approved):
let's goo! 🚢

PR #153, opened 2024-01-25
feat: Migrate ranger docs to v2 layout

lachlancollins:

See also: https://github.com/TanStack/ranger/pull/85


app/projects/ranger.tsx
```diff
@@ -1,6 +1,75 @@
...
+} as const
+
+export type Framework = keyof typeof frameworks
+
+export const createLogo = (version?: string) => (
+  <>
+    <Link to="/" className="font-light">
+      TanStack
+    </Link>
+    <Link to=".." className={`font-bold`}>
```
fulopkovacs: wdyt?

```suggestion
    <Link to=".." className="font-bold">
```

app/routes/ranger.$version._index.tsx
```diff
@@ -7,17 +7,17 @@ import {
...
 import { VscPreview } from 'react-icons/vsc'
 import { RiLightbulbFlashLine } from 'react-icons/ri'
-import { gradientText, v1branch } from '~/projects/ranger'
+import { gradientText, getBranch } from '~/projects/ranger'
 import { Carbon } from '~/components/Carbon'
 import { Footer } from '~/components/Footer'
 import SponsorPack from '~/components/SponsorPack'
 import { getSponsorsForSponsorPack } from '~/server/sponsors'
-import type { LoaderFunction } from '@remix-run/node'
+import type { Framework } from '~/projects/form'
```
fulopkovacs, line 20:
```suggestion
import type { Framework } from '~/projects/ranger'
```

app/routes/ranger.$.tsx
```diff
@@ -24,8 +24,7 @@ function handleRedirects(context: Parameters<LoaderFunction>[0]) {
   reactRangerV1List.forEach((item) => {
     if (url.pathname.startsWith(`/ranger/v1/${item.from}`)) {
       throw redirect(
-        `/ranger/v1/${item.to}?from=reactRangerV1&original=https://react-ranger-v1.tanstack.com/${item.from}`,
-        301
+        `/ranger/v1/${item.to}?from=reactRangerV1&original=https://react-ranger-v1.tanstack.com/${item.from}`
```
fulopkovacs, line 27: The latest version is `v0` according to `app/projects/ranger.tsx`, so shouldn't we redirect users to the `v0` docs? 🤔 (It does sound weird to go from v1 to v0, I admit that...)

fulopkovacs (changes requested):
I get an error for ranger doc routes where version is not `v0` or `latest`. The error message says that `selectedOption is undefined`.

See these routes for example:
- localhost:3000/ranger/v1/docs/overview
- localhost:3000/ranger/v8/docs/overview

image

(I am on the `docs-rework` branch from https://github.com/TanStack/tanstack.com/pull/153 in my local clone of the `tanstack/ranger` repo.)

app/routes/ranger.$version._index.tsx
lachlancollins: Oops, thanks!

app/routes/ranger.$.tsx
```diff
@@ -4,29 +4,19 @@ import { redirect } from '@remix-run/node'
...
 
   reactRangerV1List.forEach((item) => {
     if (url.pathname.startsWith(`/ranger/v1/${item.from}`)) {
-      throw redirect(
-        `/ranger/v1/${item.to}?from=reactRangerV1&original=https://react-ranger-v1.tanstack.com/${item.from}`,
-        301
-      )
+      throw redirect(`/ranger/v0/${item.to}`)
     }
   })
```
fulopkovacs: Hmmm, something is still wrong with this code, because `/ranger/v1/docs/overview` should redirect me to `/ranger/v0/docs/introduction`, but I get this screen instead:
image

lachlancollins:
> I get an error for ranger doc routes where version is not `v0` or `latest`. The error message says that `selectedOption is undefined`.
> 
> See these routes for example:
> 
> * localhost:3000/ranger/v1/docs/overview
> * localhost:3000/ranger/v8/docs/overview

I think this is a development-only problem, with a different issue in production. Since local development uses the file system, the version selector doesn't work (different git branches). If you delete the `fetchFs` function from `fetchRepoFile` in a project with the v2 layout, it instead returns something like this:

image

This is for http://localhost:3000/table/v9/docs/introduction

For now, I'm just going to change the 'latest' branch back to v1, but this will require a general fix for the new docs layout, rather than a specific fix for ranger.

I'd imagine the fix would be something like checking if `$version` was a valid entry in `availableVersions`, and if not redirecting to latest.

app/routes/ranger.$.tsx
lachlancollins: Yep, see the comment below

lachlancollins:
> I'd imagine the fix would be something like checking if `$version` was a valid entry in `availableVersions`, and if not redirecting to latest.

This was an incredibly quick fix to implement :)

fulopkovacs (approved):
Awesome, works perfectly! ☺️ 🚀

PR #154, opened 2024-01-25
feat: Migrate router docs to v2 layout

lachlancollins:

See also: https://github.com/TanStack/router/pull/1081


fulopkovacs (approved):
LTGM! 🚀

PR #155, opened 2024-01-25
fix: Invalid versions redirect to latest

lachlancollins:
Fixes the error that occurs for invalid versions (e.g. /table/v9/docs) described on #153


fulopkovacs (approved):
LGTM 🚢

PR #157, opened 2024-01-26
refactor: Simplify docs logic, fix loading framework from localStorage

lachlancollins:
- Removed `v2`, `frameworkConfig` and `versionConfig` options on `<DocsLayout />`
- Moved `createLogo` logic into `<DocsLogo />` component
- Replaced `gradientText` with `colorFrom` and `colorTo`, which are now also applied to `<DocsLayout />`
- Marked `frameworkMenus` as required field in `config.json`
- Fixed missing docs when the framework saved in localStorage isn't available for that project

### DIFF
--- a/app/utils/config.ts
+++ b/app/utils/config.ts
@@ -15,6 +15,7 @@ export type MenuItem = {
   children: {
     label: string | React.ReactNode
     to: string
+    badge?: string
   }[]
 }
 
@@ -24,7 +25,6 @@ const menuItemSchema = z.object({
     z.object({
       label: z.string(),
       to: z.string(),
-
       badge: z.string().optional(),
     })
   ),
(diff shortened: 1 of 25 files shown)


tannerlinsley (approved):

app/utils/config.ts
```diff
@@ -98,33 +98,23 @@ export const useDocsConfig = ({
   const match = matches[matches.length - 1]
   const params = useParams()
   const version = params.version!
-  const framework = localStorage.getItem('framework') || 'react'
+  const localStorageFramework = localStorage.getItem('framework')
+  const framework = localStorageFramework
```
lachlancollins, line 102: Fixes missing docs when the framework saved in localStorage isn't available for that project

PR #158, opened 2024-01-26
refactor: Clean up old layout logic

lachlancollins:
- Use shared old version banner
- Move invalid version redirect from docs level to version level (e.g. `/query/v8` will redirect to `/query/latest`)
- Move `useDocsConfig` hook into `<DocsLayout />` and refactor into 3 smaller hooks (`useMenuConfig` `useFrameworkConfig`, `useVersionConfig`)
- Moved `localMenu` out of individual project config files into `useMenuConfig`
- Replace `LoaderFunction` with `LoaderFunctionArgs`
- Replace codesandbox with stackblitz
- Get version from loader rather than `useParams`
- Fixes some eslint warnings


app/components/DocsLayout.tsx
```diff
@@ -12,27 +17,193 @@ import { useLocalStorage } from '~/utils/useLocalStorage'
...
+  const { framework: paramsFramework } = useParams()
+  const localStorageFramework = localStorage.getItem('framework')
+
+  return paramsFramework
+    ? paramsFramework
+    : localStorageFramework
+    ? frameworks[localStorageFramework]
+      ? localStorageFramework
+      : 'react'
+    : 'react'
```
fulopkovacs, line 39: wdyt? (You don't have to change this if you don't want to, it's just a suggestion!)
```suggestion
    return (
    paramsFramework ||
    (localStorageFramework && localStorageFramework in frameworks
      ? localStorageFramework
      : 'react')
    )

```

lachlancollins: Yeah that's definitely nicer 👍

app/routes/table.$.tsx
```diff
@@ -43,15 +43,9 @@ function handleRedirects(context: LoaderFunctionArgs) {
...
+  reactTableV7List.forEach((item) => {
     if (url.pathname.startsWith(`/table/v7/${item.from}`)) {
-      throw redirect(
-        `/table/v8/${
-          item.to
-        }?from=reactTableV7&original=https://github.com/TanStack/table/tree/v7/docs/src/pages/${
-          item.from
-        }${item.from ? '.md' : ''}`
-      )
+      throw redirect(`/table/v8/${item.to}?from=reactTableV7`)
```
fulopkovacs, line 48: It seems like this redirect will never, because routes like `/table/v7/docs/api/useColumnOrder` are handled in `app/routes/table.$version.docs.$.tsx` and not this file. 

This is what happens to routes like `/table/v7/docs/api/useColumnOrder`:
```sh
/table/v7/docs/api/useColumnOrder
/table/latest/docs/api/useColumnOrder #1st redirect
/table/latest/docs/introduction #2nd redirect
```

lachlancollins: Yeah that's what I thought, I originally removed these because I didn't think they did anything. What do you think about removing? I guess they could be put in the `docs.$` level?

app/routes/virtual.$.tsx
```diff
@@ -24,15 +25,9 @@ function handleRedirects(context: LoaderFunctionArgs) {
...
+  reactVirtualV2List.forEach((item) => {
     if (url.pathname.startsWith(`/virtual/v2/${item.from}`)) {
-      throw redirect(
-        `/virtual/v3/${
-          item.to
-        }?from=reactVirtualV2&original=https://github.com/TanStack/virtual/tree/v2/docs/src/pages/${
-          item.from
-        }${item.from ? '.md' : ''}`
-      )
+      throw redirect(`/virtual/v3/${item.to}?from=reactVirtualV2`)
```
fulopkovacs, line 30: This redirect is not thrown for routes that begin with `/virtual/v2/docs` (e.g.: `/virtual/v2/docs/installation`) because those are handled in `app/routes/virtual.$version.docs.$.tsx`.

There's an issue with paths beginning with `/virtual/v2/examples` too, but that wasn't introduced in this PR. Basically, 
```
/virtual/v2/examples/dynamic
``` 
redirects to
```
/virtual/v3/examples/react/dynamic?from=reactVirtualV2
```
instead of
```
/virtual/v3/docs/framework/react/examples/dynamic?from=reactVirtualV2
```

lachlancollins: Yeah - I might handle this in a separate PR if you're okay with leaving if for now?

app/routes/router.$.tsx
```diff
@@ -1,16 +1,16 @@
...
 
   return redirect('/router/latest')
 }
 
-function handleRedirects(context: Parameters<LoaderFunction>[0]) {
+function handleRedirects(context: LoaderFunctionArgs) {
   const url = new URL(context.request.url)
   // prettier-ignore
-  const reactLocationv2List = [
+  const reactLocationV2List = [
```
fulopkovacs, line 13: I think this was copied here from `app/routes/virtual.$.tsx`, because the list matches the list we have there and Router doesn't even have a v2. 

So this whole `handleRedirects` function could be deleted at this point.

lachlancollins: Oops you're right, I copied from the wrong spot!

fulopkovacs (changes requested):
I found a few issues around redirecting old doc paths. I think some of them were not introduced in this PR, so you don't have to address them here (you can just open up issues and assign them to me).

app/routes/router.$.tsx
lachlancollins: Wait no I copied the right thing back, it's just that the original wasn't functional. Again, I might fix this in a subsequent PR?

app/routes/virtual.$.tsx
fulopkovacs: Yeah, a separate PR sounds fine!

lachlancollins:
> I found a few issues around redirecting old doc paths. I think some of them were not introduced in this PR, so you don't have to address them here (you can just open up issues and assign them to me).

Yep I think we can fix those non-functional redirects in a subsequent PR (I should've just not touched those files oops) - is there anything else that needs to be fixed in this PR?

app/routes/table.$.tsx
fulopkovacs: Yeah, I think they should be there. Btw the routes are messed up anyways. See this one:
```tsx
{from: 'docs/api/useColumnOrder',to: 'docs/api/features/column-ordering',}
```
The value for the `to` property should be `docs/guide/column-ordering`. 🤣

app/routes/router.$.tsx
fulopkovacs: That's cool with me! 👍

fulopkovacs (approved):
Ok, the issues I found are not specific to this PR, so they'll be fixed separately! LGTM! 🚀

lachlancollins:
Thanks heaps @fulopkovacs !

PR #162, opened 2024-01-26
Try fixing algolia search for Query again

fulopkovacs:
https://github.com/TanStack/tanstack.com/pull/160 fixed this route:
- https://tanstack.com/query/latest/docs/react/reference/QueryClient
but broke this one:
- https://tanstack.com/query/latest/docs/react/reference/hydration

When I was making that PR I was unaware of the fact that not all `reference` docs are in same directory in `tanstack/query`. We currently have three directories named "reference":
- docs/framework/react/reference
- docs/framework/vue/reference
- docs/reference

### DIFF
--- a/app/routes/query.$version.docs.framework.$framework.$.tsx
+++ b/app/routes/query.$version.docs.framework.$framework.$.tsx
@@ -10,12 +10,21 @@ export const loader = async (context: LoaderFunctionArgs) => {
   const { '*': docsPath, framework, version } = context.params
   const { url } = context.request
 
+  let redirectPath = url.replace(/\/docs.*/, '/docs/framework/react/overview')
+
+  if (url.includes(`docs/framework/${framework}/reference`)) {
+    redirectPath = url.replace(
+      `docs/framework/${framework}/reference`,
+      'docs/reference'
+    )
+  }
+
   return loadDocs({
     repo,
     branch: getBranch(version),
     docPath: `docs/framework/${framework}/${docsPath}`,
     currentPath: url,
-    redirectPath: url.replace(/\/docs.*/, '/docs/framework/react/overview'),
+    redirectPath,
   })
 }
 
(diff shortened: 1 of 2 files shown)


app/routes/query.$version.docs.framework.$framework.$.tsx
```diff
@@ -10,12 +10,21 @@ export const loader = async (context: LoaderFunctionArgs) => {
...
+      'docs/reference'
+    )
+  }
+
   return loadDocs({
     repo,
     branch: getBranch(version),
     docPath: `docs/framework/${framework}/${docsPath}`,
     currentPath: url,
-    redirectPath: url.replace(/\/docs.*/, '/docs/framework/react/overview'),
```
lachlancollins, line 18: I think we could just replace this with  `url.replace(`/framework/${framework}/`, '')` - if any framework-specific route fails to match, it would try redirecting to the core equivalent, which would then redirect to the overview doc from that loader if that also failed.

lachlancollins (changes requested):
What do you think of this idea?

app/routes/query.$version.docs.framework.$framework.$.tsx
lachlancollins: Sorry that code should actually be:
```
url.replace(`/framework/${framework}/`, '/')
```

lachlancollins (approved):
AMAZING 🚀

app/routes/query.$version.docs.framework.$framework.$.tsx
fulopkovacs: I like it. Tested it locally, and it seems to work!

lachlancollins (changes requested):
R.e. potential for infinite redirects, can we also change the core-level redirect from `'/docs/framework/react/overview'` to `'/docs/overview'`? I'll create a PR to query.

lachlancollins (approved):
Sorry scrap that for now, let's just merge this, it works

PR #165, opened 2024-01-28
feat: Add routes for TanStack Config

lachlancollins:
https://github.com/TanStack/config

### DIFF
--- a/app/projects/config.ts
+++ b/app/projects/config.ts
@@ -1,0 +1,23 @@
+import { FaDiscord, FaGithub } from 'react-icons/fa'
+import { useDocsConfig } from '~/utils/config'
+import type { ConfigSchema, MenuItem } from '~/utils/config'
+
+export const repo = 'tanstack/config'
+
+export const latestBranch = 'main'
+export const latestVersion = 'v0'
+export const availableVersions = ['v0']
+
+export const colorFrom = 'from-gray-500'
+export const colorTo = 'to-gray-700'
+export const textColor = 'text-gray-700'
+
+export const frameworks = {} as const
+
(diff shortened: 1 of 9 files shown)


app/projects/config.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const colorFrom = 'from-gray-500'
+export const colorTo = 'to-gray-700'
+export const textColor = 'text-gray-700'
+
+export const frameworks = {
+  react: { label: 'React', logo: reactLogo, value: 'react' },
+  solid: { label: 'Solid', logo: solidLogo, value: 'solid' },
+  vue: { label: 'Vue', logo: vueLogo, value: 'vue' },
+} as const
```
crutchcorn, line 19: We don't have any frameworks 😅

crutchcorn (changes requested):
Code looks good minus this one thing.

app/projects/config.ts
lachlancollins: Yeah, this annoys me too - unfortunately the logic in `<DocsLayout />` currently requires the frameworks object to have items, or it crashes. I was hoping to refactor this in a separate PR.

Since React, Solid and Vue are frameworks we at least recommend using Config's build with, I figured it was a suitable compromise for now. Lmk if you'd prefer we fix this first!

crutchcorn: Let's try to fix first, otherwise I think it'll lead to a confusing UX for new users trying to figure out what config was

lachlancollins: Fixed in https://github.com/TanStack/tanstack.com/pull/186

Issue #164, opened 2024-01-28
Old manual version redirects don't work as expected


lachlancollins:
## Problem

The `handleRedirects` functions in `table.$.tsx`, `router.$.tsx`, `query.$.tsx`, and `virtual.$.tsx` do not work as expected, since a lower-level layout has priority on redirecting these paths.

For example, if you visit `/table/v7/docs/api/useColumnOrder`:

```
/table/v7/docs/api/useColumnOrder
/table/latest/docs/api/useColumnOrder #1st redirect
/table/latest/docs/introduction #2nd redirect
```

See discussion on #158 

## Potential Solution

The `handleRedirects` functions should probably be moved to the `docs.$.tsx` level, where they can run before the `loadDocs` function. Maybe the original functions could be moved to and exported from the project folders?

fulopkovacs:
> The handleRedirects functions should probably be moved to the docs.$.tsx level, where they can run before the loadDocs function. Maybe the original functions could be moved to and exported from the project folders?

Good ideas, I'll make a PR!

PR #170, opened 2024-01-31
Fix old manual redirects

fulopkovacs:
Fixes the old manual version redirects don't work as expected  (https://github.com/TanStack/tanstack.com/issues/164).

PR #173, opened 2024-02-04
Fix the old manual redirects for Virtual and Query

fulopkovacs:
This PR was the last one we needed to fix https://github.com/TanStack/tanstack.com/issues/164 fully.

| project | PR |
| --  | -- |
| `query` | this PR |
| `virtual` | this PR |
| `router` | https://github.com/TanStack/tanstack.com/pull/170  (*) |
| `table` | https://github.com/TanStack/tanstack.com/pull/170 |

> (*) **About `router`**: I only removed the manual redirects here, because I was unable to find the correct routes. The oldest commit (e4c665c) I could found containing the `"from=reactLocationV2"` string had the virtual routes in it, which seemed to be an error (see `git log -G 'from=reactLocationV2' --oneline -p`).


lachlancollins (approved):
Amazing 🚀

PR #182, opened 2024-02-09
Update stackblitz embed src from /simple → /fixed

### LINKED ISSUE #174: preview broken at `tanstack.com/virtual`
currently seeing this error at `https://tanstack.com/virtual/latest`

image

acusti:
fixes #174

- `examples/${framework}/simple` doesn’t exist in any of the `react`, `svelte`, or `vue` examples folders
- `examples/${framework}/fixed` does exist in all of the framework example folders and seems like the equivalent to “simple”

### DIFF
--- a/app/routes/ranger.$version._index.tsx
+++ b/app/routes/ranger.$version._index.tsx
@@ -317,7 +317,7 @@ export default function TanStackRouterRoute() {
       <div className="bg-white dark:bg-black">
         <iframe
           key={framework}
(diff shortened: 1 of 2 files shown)


fulopkovacs (changes requested):
Thanks for the fix! ☺️ Could you fix the same issue for Ranger too? The StackBlitz project should point at the `basic` example (https://github.com/TanStack/ranger/tree/main/examples/react/basic) instead of `simple` (which does not exist).

acusti:
@fulopkovacs thanks for the quick review! i updated the PR as requested.

fulopkovacs (approved):
Awesome, thanks for the contribution! 🚀

PR #186, opened 2024-02-16
refactor: `<Select />` takes array of options rather than object of objects

lachlancollins:
The `available` field now uses type `Array<{ label: string; value: string; logo?: string}>` rather than `Record<string, { label: string; value: string; logo?: string }>`

This was the most pain-free way to allow a project to have no frameworks (i.e. tanstack/config will be able to `export const frameworks = []`, and no framework selector will appear).


fulopkovacs (approved):
LGTM! 🚀

Issue #191, opened 2024-03-12
The lastest url in banner don't update after navigation


chigia001:
When working on adding canonical tag for doc site, I try to use loader result from tanstack.com/app/routes/query.$version.tsx to generate the canonical link.

This work fine for SSR inital response, but when user naivate to different page using the menu the redirect URL is not update.

In context of adding canonical link, this is not an issue as we only need this for bot.

But I believe the version banner should be up to date with the current url.

SeanCassiere:
Closed the linked PR with:

PR #192, opened 2024-03-12
Add canonical link to tanstack/query page

chigia001:
There are a know issue with this approach:
https://github.com/TanStack/tanstack.com/issues/191

But alternative solution mean modify each route to capture the request and generate canonical url there.


SeanCassiere (approved):

SeanCassiere:
This change should no longer be needed _I think_, since Router should be accumulating and making the meta changes on navigation.

If not, please ping the original issue and we can look at getting a PR pushed through.

PR #204, opened 2024-04-06
fix: dark mode for the docsearch component

SeanCassiere:
Restores the previous functionality of having the DocSearch component respond to the user's color-schema preference.

Also, added a function to the script to watch for changes to the preference, and update the `data-theme` attribute accordingly.

Something might have changed from our previous implementation, but now DocSearch's default styles only recognize the `data-theme` attribute when it's being set on the HTML document (and not the body), so these changes reflect that.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,4 @@
 **/build
 **/public
 pnpm-lock.yaml
+routeTree.gen.ts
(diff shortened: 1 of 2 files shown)


TkDodo (approved):

.prettierignore
```diff
@@ -2,3 +2,4 @@
 **/build
 **/public
 pnpm-lock.yaml
+routeTree.gen.ts
```
SeanCassiere, line 5: CI was failing due to formatting on the routeTree.gen.ts. So I excluded the file from prettier format, since the file is handled by TSR.

PR #227, opened 2024-05-14
feat(router): opt-out of codesanbox examples

SeanCassiere:
Opting-out of providing codesanbox urls for Router.

CSB repros are practically impossible to make changes to and test alternatives **AND** they make you need to pray to a deity to actually get the packages to upgrade to latest.


KevinVandy:
failing build, but also, is completely hiding CSB a good idea? I've dealt with the issues above, but it can still be a good alternative for some.

KevinVandy (approved):

PR #244, opened 2024-06-17
fix: headings collapsing when there is no content to separate them

SeanCassiere:
Currently, all headings extracted from the markdown files are wrapped in anchor tags.

Since anchor tags have a display property of `inline`, when there are adjacent headings and no content to separate them, it causes the headings to collapse onto one another into a single line. See the screenshot below.

Screenshot 2024-06-17 at 20 29 05

The easiest fix for this, without having to use specific CSS selectors, is to set the `display: grid;` on the outer block containing the markdown content. This fixes the bug and restores order to world 🙌🏼. See the screenshot below.

Screenshot 2024-06-17 at 21 53 39

### DIFF
--- a/app/components/Doc.tsx
+++ b/app/components/Doc.tsx
@@ -21,7 +21,7 @@ export function Doc({
       <div className="h-4" />
       <div className="h-px bg-gray-500 opacity-20" />
       <div className="h-4" />
-      <div className="prose prose-gray prose-sm dark:prose-invert max-w-none grid">
+      <div className="prose prose-gray prose-sm dark:prose-invert max-w-none md:grid">
         <Markdown code={content} />
       </div>
       <div className="h-12" />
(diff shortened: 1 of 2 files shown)


app/components/Doc.tsx
```diff
@@ -21,7 +21,7 @@ export function Doc({
       <div className="h-4" />
       <div className="h-px bg-gray-500 opacity-20" />
       <div className="h-4" />
-      <div className="prose prose-gray prose-sm dark:prose-invert max-w-none">
+      <div className="prose prose-gray prose-sm dark:prose-invert max-w-none md:grid">
```
SeanCassiere, line 24: Using a breakpoint here, since the existing styling for mobile needs it to not bork things.

KevinVandy (approved):

PR #249, opened 2024-07-08
feat: adding support for gfm alerts

SeanCassiere:
Adding UI support for GFM Alerts.

These GFM Alert types are supported. Below is the output of Github's Alerts.

> [!NOTE]
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Highlights information that users should take into account, even when skimming.

> [!IMPORTANT]
> Highlights information that users should take into account, even when skimming.

> [!WARNING]
> Highlights information that users should take into account, even when skimming.

> [!CAUTION]
> Highlights information that users should take into account, even when skimming.

This is the input required to insert a GFM Alert. The options are `NOTE`, `TIP`, `IMPORTANT`, `WARNING`, and `CAUTION`.

```markdown
> [!NOTE]
> Highlights information that users should take into account, even when skimming.
``` 

This is what it'd look like on the docs pages.

- Light mode.

image

- Dark mode.

image


app/styles/app.css
```diff
@@ -461,3 +461,56 @@ pre .logger.log-log svg {
   @apply translate-x-[10px] translate-y-[10px];
   @apply dark:[color-scheme:dark];
 }
+
+/* Markdown Alerts */
+.markdown-alert {
+  @apply border-l-4 pl-4 py-1.5 my-2 dark:bg-gray-900 rounded-r-sm;
```
SeanCassiere, line 467: The background on when in dark mode is not something Github does. I chose to add it in so it better stood out on the page since our page's dark background makes it all blend in like regular page content.

SeanCassiere (commented):
It pretty much matches what's being done by Github with the exception of;

-  Adding opacity to the border - This is to match what we currently have going on the website for blockquotes.
- Adding a background to alert in dark mode - This is to make the alert content stand out  more on the page.

tobySolutions (approved):
Amazing stuff! LGTM! Will be super helpful for our work on Tanstack forms' docs.

KevinVandy (approved):

PR #251, opened 2024-07-11
feat: add Lit to Virtual

kadoshms:
Add Link to `@tanstack/lit-virtual`.

### DIFF
--- a/app/routes/_libraries.virtual.$version.index.tsx
+++ b/app/routes/_libraries.virtual.$version.index.tsx
@@ -396,6 +396,7 @@ export default function RouteComp() {
           <div className="flex flex-wrap gap-2 justify-center">
             {(
               [
+                { label: 'Lit', value: 'lit' },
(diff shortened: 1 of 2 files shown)


app/routes/_libraries.virtual.$version.index.tsx
```diff
@@ -396,6 +396,7 @@ export default function RouteComp() {
           <div className="flex flex-wrap gap-2 justify-center">
             {(
               [
+                { label: 'Lit', value: 'lit' },
```
SeanCassiere, line 399: We should probably follow the same ordering as Table and place the label in the right position in the dropdown.

https://tanstack.com/table/latest/docs/introduction

kadoshms: No problem, fixed it, first attempt was simply alphabetically but that makes sense

SeanCassiere (approved):

PR #255, opened 2024-07-18
fix: Adjust markdown links to correct level

lachlancollins:
## Problem

Currently, markdown links are written in a way that is compatible with tanstack.com, but not with native markdown navigation.

For example: https://github.com/TanStack/query/blob/13817e953743537ffb9aab4da174583055be4d81/docs/framework/react/overview.md?plain=1#L99

Clicking either of these links will fail if you navigate on GitHub or in your IDE.

## Solution

If we write the docs using correct markdown links, we can use regex on tanstack.com to correct the path:

```js
const relativeHref = props.href?.replace(/([A-Za-z][A-Za-z/_-]+).md/, '../$1')
```

- `([A-Za-z][A-Za-z/_-]+)` matches group 1 starting with a letter, then optionally also containing `/`, `-` or `_`
- `.md` matches the markdown extension outside the group (**existing links will NOT break as they do not have `.md`**)
- `../$1` prefixes group 1 with `../`

For example, this rewrites `reference/FormApi.md` to `../reference/FormApi`

This even works for something like `./reference/FormApi.md`, which gets rewritten to `./../reference/FormApi` but still correctly routes on tanstack.com.


lachlancollins:
@crutchcorn this also fixes links generated by TypeDoc: https://tanstack-com-git-markdown-links-tanstack.vercel.app/form/latest/docs/reference/index

TkDodo (approved):

PR #270, opened 2024-09-10
fix: use absolute url for og:image tag

Balastrong:
Sharing https://tanstack.com/ in social networks (like linkedin) does not properly show the og:image in the preview.

<details><summary>See images</summary>
<p>

image

image

</p>
</details> 

Our current value is `/_build/assets/og-pGgYlhyc.png` but it seems [only absolute urls are valid](https://stackoverflow.com/a/9858694).

I'm not sure this works as I cannot reliably test it locally but I'd probably give it a try.
I think other pages may be affected but let's do it step by step (or if anyone has a better solution is welcome!)

### DIFF
--- a/app/routes/__root.tsx
+++ b/app/routes/__root.tsx
@@ -40,7 +40,7 @@ export const Route = createRootRouteWithContext<{
     ...seo({
       title: 'TanStack | High Quality Open-Source Software for Web Developers',
       description: `Headless, type-safe, powerful utilities for complex workflows like Data Management, Data Visualization, Charts, Tables, and UI Components.`,
-      image: ogImage,
+      image: `https://tanstack.com${ogImage}`,
       keywords:
         'tanstack,react,reactjs,react query,react table,open source,open source software,oss,software',
     }),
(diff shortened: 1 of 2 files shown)


SeanCassiere (approved):

SeanCassiere (commented):
On second though, maybe we should be using one of Vercel's publish URL environment variables?
https://vercel.com/docs/projects/environment-variables/framework-environment-variables

Balastrong:
> On second though, maybe we should be using one of Vercel's publish URL environment variables? https://vercel.com/docs/projects/environment-variables/framework-environment-variables

Agree, that would be probably more correct!
I didn't find any of those used in the codebase so I went with the hardcoded url to at least validate the fix.

Side note, do we even have those variable available? I mean `NEXT_XXX` while we're on a TanStack Start project.

SeanCassiere:
> Side note, do we even have those variable available? I mean `NEXT_XXX` while we're on a TanStack Start project.

Since Start uses Vinxi, you should be able to use the same variables used for SolidStart and Vite. `VITE_VERCEL_URL` is probably the variable you are looking for.

Balastrong:
Done!

I also changed the `name` attribute to `property` for og tags as [that's what meta recommends](https://developers.facebook.com/docs/sharing/webmasters/?locale=en_US) and in fact I also noticed something was missing in title/description when sharing the url so that may be one more needed fix.

Let's see what happens in the vercel preview :D

SeanCassiere (approved):

PR #271, opened 2024-09-15
docs: add blog post for ts performance of router

chorobin:


tannerlinsley:
What a great blog post. Seriously so good.

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+    LayoutLayout2Route: LayoutLayout2Route.addChildren({
+      LayoutLayout2LayoutARoute,
+      LayoutLayout2LayoutBRoute,
+    }),
+  }),
+  PostsRoute: PostsRoute.addChildren({ PostsPostIdRoute, PostsIndexRoute }),
+})
+```
+
+Generating the route tree came as a consequence of reducing tedious configuration of a route tree but keeping inference where it matters. This is where the first important change is introduced leading to better editor performance. Instead of inferring the route tree, we can take advantage of this generation step to _declare the route tree_.
```
SeanCassiere, line 36:
```suggestion
This generation process was initially introduced to reduce the tedious complexity involved in configuring your route tree, whilst still keeping inference where it mattered. This is where the first important change was introduced. Instead of inferring the route tree in a single shot, we can take advantage of this generation step to _declare the route tree_.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+const rootRouteChildren: RootRouteChildren = {
+  IndexRoute: IndexRoute,
+  LayoutRoute: LayoutRouteWithChildren,
+  PostsRoute: PostsRouteWithChildren,
+}
+
+export const routeTree = rootRoute._addFileChildren(rootRouteChildren)
+```
+
+Note the use of an `interface` to declare the children to compose the route tree. This process is repeated for all routes and children to create the route tree. After this change, the trace also changed that gave us more insight into what the compiler is now doing.
```
SeanCassiere, line 54:
```suggestion
Note the use of an `interface` to declare the children to compose the route tree. This process is repeated for all routes and their children when generating the route tree. With this change, running a trace gave us a much better idea of what was happening inside the language-service.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+<img src="/blog-tracing-declare-route-tree.png" />
+
+This is still slow and we're not quite there yet but there is something - _the trace is different_. The inference for the whole route tree is still happening but somewhere else. After working through the types, it turned out to be a type called `ParseRoute`
```
SeanCassiere, line 58:
```suggestion
This is still slow and we're not quite there yet but there is something - _the trace is different_. The type inference for the entire route tree was still happening, but it was now being done _somewhere else_. After working through our types, it turned out to be happening in a type named `ParseRoute`.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+}
+  ? unknown extends TChildren
+    ? TAcc
+    : TChildren extends ReadonlyArray<any>
+    ? ParseRoute<TChildren[number], TAcc | TChildren[number]>
+    : ParseRoute<TChildren[keyof TChildren], TAcc | TChildren[keyof TChildren]>
+  : TAcc
+```
+
+This type walks down the route tree to create a union of all routes. The union in turn is used to create a type mapping from `id` -> `Route`, `from` -> `Route` and also `to` -> `Route`. An example of this mapping exists as a mapped type
```
SeanCassiere, line 72:
```suggestion
This type walks down the route tree to create a union of all routes. The union, in turn, is used to create a type mapping from `id` -> `Route`, `from` -> `Route` and also `to` -> `Route`. An example of this mapping exists as a mapped type.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+```tsx
+export type RoutesByPath<TRouteTree extends AnyRoute> = {
+  [K in ParseRoute<TRouteTree> as K['fullPath']]: K
+}
+```
+
+The important realization here is for file based routing we generate the route tree and therefore we don't need to use `ParseRoute` and we can skip this work by generating the mapping type when generating the route tree. Instead we can generate the following:
```
SeanCassiere, line 80:
```suggestion
The important realization here was that when using file-based routing, we were able to skip the `ParseRoute` type entirely by outputting that mapping type ourselves whenever the route tree was generated. Instead, we'd be able to generate the following:
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+}
+
+export const routeTree = rootRoute
+  ._addFileChildren(rootRouteChildren)
+  ._addFileTypes<FileRouteTypes>()
+```
+
+In addition to declaring children, we also declare interfaces which map paths to a route.
+
+This change along with other type level changes to conditionally use `ParseRoute` only if these types are not registered resulted in a trace which was our aim all along 🥳
```
SeanCassiere, line 155:
```suggestion
This change, along with other type-level changes to conditionally use the `ParseRoute` type only if these types were not registered, resulted in a trace which was our aim all along 🥳.
```

The ending `resulted in a trace which was our aim all along 🥳` could probably use some work.

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+<img src="/blog-tracing-faster.png">
+
+The first file to reference a `Link` no longer triggers inference from the whole route tree which increases perceived language service speed significantly
```
SeanCassiere, line 159:
```suggestion
The first file referencing a `<Link>` no longer triggers a comprehensive route tree inference, thereby significantly boosting the user's perceived language service speed.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+By doing this TypeScript will only infer the types required for a specific route when it is referenced by a `Link`. This may not translate to overall better typescript check time if all routes are used but is significant for the language service
```
SeanCassiere, line 161:
```suggestion
By doing so, TypeScript will only infer the types required for a specific route when it is referenced by a `<Link>`. This may not translate to overall better TypeScript type-checking time when all routes are being linked to, but it is a significant speed increase for the language service when in a file/region.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+The difference between the two is striking with large route trees with complex inference (400 in this example)
```
SeanCassiere, line 163:
```suggestion
The difference between the two is striking, as seen in these large route trees with complex inference (400 in this example below 🤯):
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+<div style="display: flex">
+  <video src="/language-service-slow.mp4" width="640" height="480" autoplay muted loop></video>
+  <video src="/language-service-fast.mp4" width="640" height="480" autoplay muted loop></video>
+</div>
+
+You maybe thinking that this is cheating, we are indeed making use of a generation step. However this generation step for file-based (and now virtual file based routing) was always there and a necessary step whenever you modify or create a new route. However once the route is created and the route tree is generated, inference remains throughout anything in a route definition. That means if you make any changes to `validateSearch`, `beforeLoad`, `loader` and others the type is inferred and reflected instantly. The DX in this regard has not changed but the performance in the editor can feel awesome for large route trees
```
SeanCassiere, line 170:
```suggestion
You may be thinking that this is _cheating_ since we are doing a lot of the heavy lifting here in the route tree generation phase. Our response to that is that this generation step for file-based routing (and now virtual file-based routing) was already there and was always a necessary step whenever you modified or created a new route.

So, once a route has been created and the route tree is generated, the inference remains the same throughout everything within the route definition. This means that you can make changes to `validateSearch`, `beforeLoad`, `loader`, and others, with the inferred types always being reflected instantly.

The DX has not changed, but the performance in your editor feels awesome (especially when you are working with large route trees).
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+## The ground rules
+
+This change has made a lot of the surface of TanStack Router's exports performant for the language service but there are existing areas which will still fallback to forcing inference on the whole route tree. These areas are a kind of loose or non strict modes.
```
SeanCassiere, line 174:
```suggestion
This change involved a lot of TanStack Router's exports being improved to make it more performant to consume these generated types whilst still being able to fall back on the whole route tree inference when using code-based routing. We've also still got areas of the codebase that still rely on the full route tree inference. These areas are our version of a loose/non-strict mode.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+```tsx
+<Link to="." search={{ page: 0 }} />
+<Link to=".." search={{page: 0}} />
+<Link to="/dashboard" search={prev => ({..prev, page: 0 })} />
+```
+
+All three usages of `Link` require inference of the whole route tree and are therefore falling back to a worse editor experience when interacting with them. The first two, TanStack Router does not know what route you want to navigate to and therefore tries its best to guess a very loose type from all routes. The third uses `prev` but TanStack Router does not know what `Route` you are navigating `from` and again tries to guess `prev` with a loose type from all routes. The most performant for your editor is the following:
```
SeanCassiere, line 182:
```suggestion
All three above usages of `<Link>`, require the inference of the whole route tree and therefore result in a worse editor experience when interacting with them.

In the first two instances, TanStack Router does not know what route you want to navigate to, therefore it tries its best to guess at a very loose type inferred from all the routes in your route tree. The third instance of `<Link>` from above, uses the `prev` argument in the `search` updater function, but in this instance, TanStack Router does not know which `Route` you are navigating `from`, whereby it needs to once again try to guess the loose type of `prev` by scanning the whole route tree.

The most performant usage of `<Link>` in your editor would be the following:
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+```tsx
+<Link from="/dashboard" search={{ page: 0 }} />
+<Link from="/dashboard" to=".." search={{page: 0}} />
+<Link from="/users" to="/dashboard" search={prev => ({...prev, page: 0 })} />
+```
+
+TanStack Router can narrow the types to specific routes in these cases. This means that you get better type safety and also better editor performance as your application scales. Therefore we encourage the use of `from` or `to` in these cases. To be clear `from` is only necessary if `prev` is used, otherwise TanStack Router does not infer the whole route tree
```
SeanCassiere, line 190:
```suggestion
TanStack Router can narrow the types to specific routes in these cases. This means that you get better type safety and better editor performance as your application scales. Therefore, we encourage the use of `from` and/or `to` in these cases. To be clear, in the third example, the usage of `from` is only necessary if the `prev` argument is used, otherwise, TanStack Router does not need to infer the whole route tree.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+
+```tsx
+const search = useSearch({ strict: false })
+const params = useParams({ strict: false })
+const context = useRouteContext({ strict: false })
+const loaderData = useLoaderData({ strict: false })
+const match = useMatch({ strict: false })
+```
+
+In which case for better editor performance and type safety, using `from` is recommended
```
SeanCassiere, line 202:
```suggestion
In this case, better editor performance and type safety can be achieved by using the recommended `from` property.
```

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
...
+const context = useRouteContext({ from: '/dashboard' })
+const loaderData = useLoaderData({ from: '/dashboard' })
+const match = useMatch({ from: '/dashboard' })
+```
+
+## Moving forward
+
+Going forward we believe TanStack Router is very well positioned to have the best balance between type safety and typescript performance without compromising on inference used throughout route definitions in file (and virtual file) based routing. Everything in your route definitions remains inferred but what gets generated for you is a route tree which helps the language service through inference by declaring types where it matters, something that you would never want to write yourself. This approach also appears scalable for the language service, we were able to create thousands of route definitions and the language service will remain stable provided you keep to the `strict` parts of TanStack Router. We will continue to keep improving TypeScript performance on TanStack Router to reduce overall check time and improve the language service performance further but we think this is an important milestone to share and something we hope users of TanStack Router will appreciate
```
SeanCassiere, line 214:
```suggestion
Going forward, we believe that TanStack Router is very well positioned to have the best balance between type safety and TypeScript performance, without having to compromise on the quality of the type inference used throughout the route definitions in file-based (and virtual file-based) routing. Everything in your route definitions remains inferred, with the changes in the generated route tree only aiding the language service by declaring the necessary types where it matters, something that you would never want to write yourself.

This approach also looks to be scalable for the language service. We were able to create thousands of route definitions with the language service remaining stable provided you keep to the `strict` parts of TanStack Router.

We will continue to keep improving TypeScript performance on TanStack Router to reduce overall check time and improve the language service performance further, but still felt that this was an important milestone to share and something we hope the users of TanStack Router will appreciate.
```

SeanCassiere (commented):
Perhaps the videos can be uploaded to YouTube (preferably @tannerlinsley's account marked as unlisted) and then have them be added in as embeds. No need to spend the extra bandwidth serving video 😬

app/blog/tanstack-router-typescript-performance.md
```diff
@@ -0,0 +1,214 @@
+---
+title: A milestone for TypeScript Performance in TanStack Router
```
SeanCassiere, line 2: Please add a published date, like with other articles.

chorobin: Will when we decide on the date to publish ;)

SeanCassiere (approved):

PR #304, opened 2024-11-20
fix: update getting started route in tanstack start landing page

tannerbaum:
I haven't touched these repos before or had the pnpm setup to test this, but thought I would shoot my shot at correcting the route causing a 500 based on the other links on the page 😆

### DIFF
--- a/app/routes/_libraries.start.$version.index.tsx
+++ b/app/routes/_libraries.start.$version.index.tsx
@@ -172,7 +172,7 @@ export default function VersionIndex() {
           provider.
         </p>
         <Link
-          to="./docs/framework/react/start/overview"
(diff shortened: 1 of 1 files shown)


app/routes/_libraries.start.$version.index.tsx
```diff
@@ -172,7 +172,7 @@ export default function VersionIndex() {
           provider.
         </p>
         <Link
-          to="./docs/framework/react/start/overview"
+          to="../../router/latest/docs/framework/react/start/getting-started"
```
tannerbaum, line 175: Commit suggestion for the overview page if you want that instead. For your convenience 😆 
```suggestion
          to="../../router/latest/docs/framework/react/start/overview"
```

PR #309, opened 2024-11-26
Use fresh data w/ Convex for open source stats

erquhart:
https://github.com/user-attachments/assets/61d1d7cb-e3d3-4851-9fad-ec90b1750f4f

This PR makes the OSS stats data as live as it can get:
- GitHub stars: 100% live
- Npm downloads: source data is only updated every 24h, so we use an expected amount based on available past data and run an animated counter toward that number between updates
- GitHub contributors: this data is guaranteed to be hours stale in GitHub's own api docs, and imo doesn't change enough to warrant faking between updates
- GitHub dependencies: this can only be scraped from the website. Because there's no api based historical data we're not animating change here, but we could.

Everything is updated to latest available hourly.

**Primary changes**
- Adds Convex to the project so Convex data is generally accessible to the app
- Because OSS data like this is not a unique requirement for tanstack.com, the logic and sync stuff is all in a new OSS Stats Convex component (so it's a dependency instead of being inline in the repo)
- Note: component is temporarily published on my personal npm org

**Ancillary changes**
- Changed library config files from using `satisfies` to explicit types to address errors in the libraries index component.

**Required steps**
- A maintainer should run `pnpm dev:convex` locally and create a new Convex project (you'll be prompted)
- Create a GitHub personal access token (only needs read for public repos) and add to Convex development and production deployments as `GITHUB_ACCESS_TOKEN`
- Create a webhook on the TanStack GitHub org: https://github.com/organizations/tanstack/settings/hooks/new
  - Get the HTTP Actions URL from your production Convex deployment: https://dashboard.convex.dev > Production project deployment > Settings > URL & Deploy Key > HTTP Actions URL
  - Payload URL: `<http-actions-url>/events/github`
  - Content type: `application/json`
  - Generate a secret, add in webhook creation form and in Convex production env vars as `GITHUB_WEBHOOK_SECRET`
  - Which events? > Select individual > Stars only
- The `GITHUB_ACCESS_TOKEN` and `GITHUB_WEBHOOK_SECRET` env vars should be added to both development and production Convex deployments in the dashboard
- To kick off the initial sync and begin running hourly, from the tanstack.com directory run:
  - `npx convex run stats:sync`
  - `npx convex run stats:sync --prod`
- Do step 4 in the docs for deploying to Vercel (this PR updates vercel.json so prior steps can be skipped): https://docs.convex.dev/production/hosting/vercel
- There is a hardcoded url to a Convex production instance in `app/router.tsx` - replace this with the url to the production instance that was just created. Hardcoding this allows local development to work without requiring a convex dev instance for the developer. They'll only need one if they're developing in the convex directory.

### DIFF
--- a/app/routes/_libraries/index.tsx
+++ b/app/routes/_libraries/index.tsx
@@ -132,28 +132,38 @@ const OssStats = () => {
   return (
     <div>
       <div className="p-8 grid grid-cols-1 sm:grid-cols-2 xl:grid-cols-4 gap-8 items-center justify-center place-items-center bg-white/50 dark:bg-gray-700/30 dark:shadow-none rounded-xl shadow-xl">
-        <div className="flex gap-4 items-center">
-          <FaDownload className="text-2xl" />
+        <a
+          href="https://www.npmjs.com/org/tanstack"
+          target="_blank"
+          rel="noreferrer"
+          className="group flex gap-4 items-center"
+        >
+          <FaDownload className="text-2xl group-hover:text-emerald-500 transition-colors duration-200" />
           <div>
-            <div className="text-2xl font-bold opacity-80 relative">
+            <div className="text-2xl font-bold opacity-80 relative group-hover:text-emerald-500 transition-colors duration-200">
               <NpmDownloadCounter npmData={npm} />
             </div>
-            <div className="text-sm opacity-50 font-medium italic">
+            <div className="text-sm opacity-50 font-medium italic group-hover:text-emerald-500 transition-colors duration-200">
               NPM Downloads
             </div>
           </div>
-        </div>
-        <div className="flex gap-4 items-center">
-          <FaStar className="text-2xl" />
+        </a>
+        <a
+          href="https://github.com/orgs/TanStack/repositories?q=sort:stars"
+          target="_blank"
+          rel="noreferrer"
+          className="group flex gap-4 items-center"
+        >
+          <FaStar className="group-hover:text-yellow-500 text-2xl transition-colors duration-200" />
           <div>
-            <div className="text-2xl font-bold opacity-80 leading-none">
+            <div className="text-2xl font-bold opacity-80 leading-none group-hover:text-yellow-500 transition-colors duration-200">
               <NumberFlow value={github?.starCount} />
             </div>
-            <div className="text-sm opacity-50 font-medium italic -mt-1">
+            <div className="text-sm opacity-50 font-medium italic -mt-1 group-hover:text-yellow-500 transition-colors duration-200">
               Stars on Github
             </div>
           </div>
-        </div>
+        </a>
         <div className="flex gap-4 items-center">
           <FaUsers className="text-2xl" />
           <div className="">


app/libraries/config.ts
```diff
@@ -19,4 +19,4 @@ export const configProject = {
   colorTo: 'to-gray-700',
   textColor: 'text-gray-700',
   frameworks: [],
-} satisfies Library
```
thomasballinger, line 22: any reason for this change? it means these don't get the literal types, they get the broader type of `Library` and types like "union of all IDs" can't be built out of them

app/router.tsx
```diff
@@ -1,22 +1,53 @@
...
-    defaultErrorComponent: DefaultCatchBoundary,
-    defaultStaleTime: 1,
-    defaultNotFoundComponent: () => {
-      return <NotFound />
-    },
-    context: {
-      assets: [],
+  const CONVEX_URL =
+    (import.meta as any).env.VITE_CONVEX_URL ||
+    'https://intent-pigeon-358.convex.cloud'
```
thomasballinger, line 13: maybe a comment about what this instance is and who owns it?

app/router.tsx
```diff
@@ -1,22 +1,53 @@
...
+  if (!CONVEX_URL) {
+    console.error('missing env var VITE_CONVEX_URL')
+  }
```
thomasballinger, line 16: these lines no longer make sense

convex/README.md
```diff
@@ -0,0 +1,90 @@
+# Welcome to your Convex functions directory!
```
thomasballinger, line 1: Maybe replace with instructions about how to develop on Convex in this package specifically

vercel.json
```diff
@@ -1,5 +1,5 @@
 {
   "framework": null,
-  "buildCommand": "pnpm build",
+  "buildCommand": "npx convex deploy --cmd 'pnpm build'",
```
thomasballinger, line 3: a conditional for "if it's a preview deploy, don't deploy the convex functions" (if this project uses Vercel preview deploys)

thomasballinger (commented):
Pretty cool!

app/libraries/config.ts
erquhart: Ah! I couldn't find a downside but that would be one. I did this because there were type errors for things like `library.badge` in the file I was working in. I'll just change it back and leave the type errors.

app/router.tsx
erquhart: I just left these in case they decide to get rid of the hardcoding approach, but I'll drop it.

app/libraries/config.ts
erquhart: Basically trying to solve this: https://github.com/microsoft/TypeScript/issues/52805

But the ruling is it's supposed to work this way, so optional properties on a type that uses satisfies have to be explicitly set to undefined. I'll do that.

vercel.json
erquhart: Is there a way you all typically go about this?

app/router.tsx
erquhart: Did this, not sure why GitHub isn't showing the change here.

vercel.json
thomasballinger: I've done `if [ "$VERCEL_ENV" = "preview" ]; then npm run build; else npx convex deploy --cmd 'npm run build'; fi` before

erquhart: I was thinking it needs to be cross platform but if it's just in vercel.json I guess not

thomasballinger:
Looks good to me! Looking forward to seeing this on the site

thomasballinger (approved):

thomasballinger:
I think x margins should match the boxes down below
image

erquhart:
@thomasballinger the current version doesn't do this, I think partly because the size of the numbers impacts the width, which changes responsively
Screenshot 2024-12-10 at 3 36 16 PM
Screenshot 2024-12-10 at 3 33 41 PM

erquhart:
Can definitely do if desired! Just pointing out that it's pre-existing.

thomasballinger:
ah gotcha!

tannerlinsley:
We're using netlify now. Does that change anything?

PR #312, opened 2024-11-29
chore: upgrade searchbox / add suggestions / fix theme

rjborba:
This PR:
- Upgrades Orama components version. It fixes several styles issue and others.
- Uses dynamic theme so SearchBox colors matches the selected theme.
- Adds chat messages suggestions

image

image

### DIFF
--- a/app/components/DocsLayout.tsx
+++ b/app/components/DocsLayout.tsx
@@ -29,6 +29,7 @@ import { DocsCalloutBytes } from '~/components/DocsCalloutBytes'
 import { ClientOnlySearchButton } from './ClientOnlySearchButton'
 import { twMerge } from 'tailwind-merge'
 import { partners } from '~/utils/partners'
+import { useThemeStore } from './ThemeToggle'
 
 // Let's use zustand to wrap the local storage logic. This way
 // we'll get subscriptions for free and we can use it in other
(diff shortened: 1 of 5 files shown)


app/components/DocsLayout.tsx
```diff
@@ -321,6 +322,10 @@ export function DocsLayout({
   const versionConfig = useVersionConfig({ versions })
   const menuConfig = useMenuConfig({ config, frameworks, repo })
 
+  const { mode: themeMode } = useThemeStore()
+
+  const oramaThemeMode = themeMode === 'auto' ? 'dark' : themeMode
+
```
tannerlinsley, line 328: You could detect auto here with a media match as done in other areas

tannerlinsley (changes requested):

app/components/DocsLayout.tsx
rjborba: Hi @tannerlinsley nice catch.
Actually Orama components does it internally when "system" is set to colorScheme.
I've updated the code so "auto" translates to "system". Is it good enough?

rjborba: @tannerlinsley, hey! Just checking if this is enough for you request. Thanks!

tannerlinsley:
Yep! Just resolve conflicts and we’re good

rjborba:
> Yep! Just resolve conflicts and we’re good

@tannerlinsley We should good to go now, then. Thanks! :)

rjborba:
@tannerlinsley Hi! Conflicts are fixed. Can we move on with this?

PR #324, opened 2024-12-28
update to @erquhart/convex-oss-stats@0.3.3

erquhart:
- improve scraping based contributor/dependent stat accuracy
- improves server and client component performance (faster sync times)


tannerlinsley:
This broke the counter

erquhart:
Fixed here: https://github.com/TanStack/tanstack.com/pull/325

PR #335, opened 2025-01-17
Rework background gradient meshes and add dots effect

laurentlahmy:
- Tries to improve background colors (where I think colors are a bit murky).
- Tries to improve background blobs parameters and colors, picked blob colors from the tanstack "island" logo.
- Adds a dot effect for texture.
- Adds a mask on screen for dark edges.

### DIFF
--- a/app/routes/__root.tsx
+++ b/app/routes/__root.tsx
@@ -185,33 +185,33 @@ function RootDocument({ children }: { children: React.ReactNode }) {
       const blobs = [
         {
           direction: [Math.random() * 1, Math.random() * 1],
-          color: { h: 184, s: 100, l: 66 }, // turquoise from logo sky (higher luminance)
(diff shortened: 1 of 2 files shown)


Willem-Jaap:
I think the dots are very noisy, especially on lower quality screens

tannerlinsley:
Could you reduce the dots in light mode a bit? They're pretty distracting, but only in light mode.

laurentlahmy:
Just pushed a commit that should fix light mode and be smarter about the desired effect.

tannerlinsley:
I think we're going to go a completely different direction actually. I have some ideas I need to explore to get the CPU usage down a lot and it's probably going to end up using more static assets/code. This was awesome and I'm really sorry I can't merge it :/ I hope you had as much fun as I did messing with canvas!

PR #340, opened 2025-02-03
refactor: Reorder TSS links

lsalling:
Use same order of library sublinks as all other libraries


lsalling:
fwiw I'd also still prefer all libraries to have the link to the docs before the examples

PR #366, opened 2025-03-03
feat: add mobile version of the "table of contents"

SeanCassiere:
Currently, on desktop, on the right side of the page, we have a persistent table of contents section that lists all the available headings within the markdown document. This section/component was hidden on mobile as its persistent nature would work on mobile.

This change, adds a mobile-optimized version of the above component, without the sticky/persistent nature.

Before:

image

After: 

image

With the section opened:

image

Also, looks surprisingly good in dark-mode as well:

image

### DIFF
--- a/app/components/DocContainer.tsx
+++ b/app/components/DocContainer.tsx
@@ -9,7 +9,7 @@ export function DocContainer({
     <div
       {...props}
       className={twMerge(
-        'w-full max-w-full space-y-2 md:space-y-6 lg:space-y-8 p-2 md:p-6 lg:p-8',
+        'w-full max-w-full p-2 md:p-6 lg:p-8',
         props.className
       )}
     >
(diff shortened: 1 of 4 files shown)


app/components/DocContainer.tsx
```diff
@@ -9,7 +9,7 @@ export function DocContainer({
     <div
       {...props}
       className={twMerge(
-        'w-full max-w-full space-y-2 md:space-y-6 lg:space-y-8 p-2 md:p-6 lg:p-8',
```
SeanCassiere, line 12: This spacing was only being used to add margin to the "Back" button on the blog page. So it was moved over to the button itself as the `space-` utils would otherwise have been applied to the`<TocMobile>` component even if it was visually hidden on the screen.

I opted not to use a React (`useMediaQuery`) hook in-favor of relying on CSS.

PR #376, opened 2025-03-22
fix(code-explorer): resolving starting path when not available

SeanCassiere:
Fixes an issue with the code-explorer dying when it couldn't resolve to child directories within the `examplePath`.

For example, when trying to start fetching the contents for `examples/react/pagination/src`, where the leaf `/src` directory did not exist, the whole application would simply fail.

To fix this, the contents will always be fetched from the contents API from the base of the `examplePath` (i.e. `examples/react/pagination`). From there on, once the contents have been fetched, the logic in `determineStartingFilePath` will resolve it to the correct starting file-path to have correct file opened in the explorer.

This results in a UI change, where the contents of the file-tree will differ from the existing implementation.

> [!IMPORTANT]
> The default opened/"starting file" will remain the same for users.

Old:
image

New: 
image

### DIFF
--- a/app/utils/documents.server.ts
+++ b/app/utils/documents.server.ts
@@ -387,8 +387,7 @@ export function fetchApiContents(
     ttl: isDev ? 1 : 1 * 60 * 1000, // 5 minute
     fn: () => {
       return isDev
-        ? // ? fetchApiContentsFs(repoPair, startingPath)
-          fetchApiContentsRemote(repoPair, branch, startingPath)
+        ? fetchApiContentsFs(repoPair, startingPath)
         : fetchApiContentsRemote(repoPair, branch, startingPath)
     },
   })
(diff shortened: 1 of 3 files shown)


app/utils/documents.server.ts
```diff
@@ -496,7 +496,7 @@ async function fetchApiContentsRemote(
     },
   }
   const res = await fetch(
-    `https://api.github.com/repos/${repo}/contents/${startingPath}?=${branch}`,
+    `https://api.github.com/repos/${repo}/contents/${startingPath}?ref=${branch}`,
```
SeanCassiere, line 499: This is what caused `index.tsx` to show up even though the actual file in the repo was `index.js`.

KevinVandy (approved):

PR #384, opened 2025-04-09
Updates

tannerlinsley:


rajivchaulagain:
@tannerlinsley  
changed from : refreshInterval: 45_000, // 45 seconds
changed to: refreshInterval: 90_000, // 45 seconds
here the comment is not changed as it should be changed from 45 to 90 seconds.

tannerlinsley:
Fixed!

PR #393, opened 2025-05-01
feat: use content collections for the blog

burnmaniac:
Based on [this X post](https://x.com/tannerlinsley/status/1918026872765727027), I've added [Content Collections](https://www.content-collections.dev/docs/quickstart/tanstack-start) to the blog page of [tanstack.com](https://tanstack.com/blog).

To summarize the change:
- Added Content Collections plugin to Vite
- Added Content Collections generated files to `.gitignore`
- Adjusted all `published` fields for articles to `YYYY-MM-DD`
- Adjusted `fetchFrontMatters` server function
- Removed `getPostList` utility function as, per my understanding, we don't need it anymore

Happy to hear feedback, as this is my small and first open-source contribution, and I've been really nervous before committing (had to fix stuff and force-push 4 times into my branch before opening PR).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,7 +15,8 @@
     "forceConsistentCasingInFileNames": true,
     "baseUrl": ".",
     "paths": {
-      "~/*": ["./app/*"]
+      "~/*": ["./app/*"],
+      "content-collections": ["./.content-collections/generated"]
     },
     "types": ["vite/client"],
     "noEmit": true
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,7 @@ yarn.lock
 dist
 .vscode/
 .env.local
+
+# Content Collections generated files
+.content-collections
+
--- a/app.config.ts
+++ b/app.config.ts
@@ -1,6 +1,6 @@
 import { sentryVitePlugin } from '@sentry/vite-plugin'
 import { defineConfig } from '@tanstack/start/config'
-import contentCollections from "@content-collections/vinxi";
+import contentCollections from '@content-collections/vite'
 import tsConfigPaths from 'vite-tsconfig-paths'
 
 export default defineConfig({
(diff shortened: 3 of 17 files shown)


tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "forceConsistentCasingInFileNames": true,
     "baseUrl": ".",
     "paths": {
-      "~/*": ["./app/*"]
+      "~/*": ["./app/*"],
+      "content-collections": ["./.content-collections/generated"]
```
SeanCassiere, line 19: Please use a more explicit alias over here. Something like `#content-collections`? or something similar?

Reading through the PR, I honestly mistook the `import { allPosts } from 'content-collections'` as a normal package.

SeanCassiere (changes requested):
First off, thank you for opening this PR!

---

My feedback 👇🏼

- What's the reasoning for changing the date format?
- Don't tie it to the Vinxi bindings
  - Since we currently working on moving TanStack Start off of Vinxi onto a Vite Plugin based setup, I'm not sure if we should commit to using the Vinxi bindings at the moment.
  - Looking at the source for both bindings, this looks like it could be possible to make this work using the Vite bindings instead.
  - Could you give it a shot using the Vite bindings? `app.config.ts` supports Vite Plugin under `vite.plugins`.

tsconfig.json
SeanCassiere: This may not be possible. Looking at the [source](https://github.com/sdorra/content-collections/blob/31664af5091fe0024bdb733c55a6810a80d43fc7/packages/vite/src/index.ts#L57) for content-collections, it looks like their Vite plugin explicitly sets it up to use `'content-collections'` are the alias.

burnmaniac:
Thank you for the feedback!

I’ve changed the date format mainly because of the schema validation, and because `YYYY-MM-DD` is generally recommended per ISO 8601. From my experience, the former format could create some issues. I think that change is reasonable.

I’m sorry about the bindings, I’ve never used Astro CC before, and I didn’t know there are Vite bindings as well.

I should have mentioned the alias cannot be changed; I tried.

SeanCassiere (approved):

SeanCassiere:
@burnmaniac thank you for contributing this!

Also, congrats on your first open-source contribution!

PR #400, opened 2025-05-07
Add i18n toggle button (en & zh-hans)

xiaoyu2er:
In this PR, we add translate redirect ability.
- add a translate button at the logo sections (`app/routes/_libraries/route.tsx` &`app/components/DocsLayout.tsx`) 
- add i18n meta links

The current translate button supports two languages: english (https://tanstack.com/) & zh-Hans (https://zh-hans.tanstack.dev/)

Here are some screenshots:

image
Google Chrome 2025-05-07 16 33 33
image
image


package.json
```diff
@@ -32,6 +32,7 @@
     "@sentry/react": "^8.35.0",
     "@sentry/vite-plugin": "^2.22.6",
     "@tailwindcss/typography": "^0.5.13",
+    "@tanstack-dev/components": "^1.4.0",
```
KevinVandy, line 35: I don't think it would be a good idea to install a 3rd party dependency like this. This component should just live in this codebase.

KevinVandy:
How do the translations work? Where do they come from? Where do they live?

KevinVandy:
I see that this PR is actually suggesting to redirect to a translated mirror on tanstack.com. While the effort is appreciated, and we do want to offer the tanstack docs in other languages, I think we need to figure out a way to do an AI translate in our build pipelines in a way in which is not too expensive. Otherwise, our ever evolving docs will get out of date in the mirrors. I almost think that it's more important for docs to be up to date than translated, as there are already browser tools that can translate our real most up to date docs, at least.

leesb971204:
@KevinVandy 

Hi, I’m a developer from Korea currently using TanStack Query and TanStack Router.
While looking for ways to help provide official TanStack documentation in other languages, I came across this PR.

In Korea, many developers are already using TanStack’s excellent libraries, and there are Korean-translated versions of the documentation created to support learning through the official docs.

I’m very interested in contributing to this effort and wanted to ask if you have any plans to officially support translations of TanStack’s documentation into multiple languages.

PR #402, opened 2025-05-12
Update site to devinxi alpha

birkskyum:
Still an issue with `createRouter is not a function` but otherwise everything should be ported


.gitignore
```diff
@@ -8,7 +8,9 @@ yarn.lock
 .vercel
 .output
 .vinxi
-.netlify
+.tanstack-start/*
```
SeanCassiere, line 11: Only ignore `.tanstack-start/build`

schiller-manuel (approved):

PR #456, opened 2025-08-06
feat: handle theme with localStorage

Balastrong:
As discussed on Discord I removed the cookie logic in favor of localStorage to handle the theme (light, dark and auto)

It previously relied on zusatand, I replaced it with a react context in a separate commit, you can see both versions in this PR.
I think we should keep it simple and stay with the context but if we want to use zustand as it's already installed I can revert the second commit.

### DIFF
--- a/src/components/ThemeProvider.tsx
+++ b/src/components/ThemeProvider.tsx
@@ -59,15 +59,18 @@ const themeDetectorScript = (function () {
   function themeFn() {
     try {
       const storedTheme = localStorage.getItem('theme') || 'auto'
+      const validTheme = ['light', 'dark', 'auto'].includes(storedTheme)
+        ? storedTheme
+        : 'auto'
 
-      if (storedTheme === 'auto') {
+      if (validTheme === 'auto') {
         const autoTheme = window.matchMedia('(prefers-color-scheme: dark)')
           .matches
           ? 'dark'
           : 'light'
         document.documentElement.classList.add(autoTheme, 'auto')
       } else {
-        document.documentElement.classList.add(storedTheme)
+        document.documentElement.classList.add(validTheme)
       }
     } catch (e) {
       const autoTheme = window.matchMedia('(prefers-color-scheme: dark)')
--- a/src/components/ThemeProvider.tsx
+++ b/src/components/ThemeProvider.tsx
@@ -1,0 +1,132 @@
+import { ScriptOnce } from '@tanstack/react-router'
+import * as React from 'react'
+import { createContext, ReactNode, useEffect, useState } from 'react'
+import { z } from 'zod'
+
+const themeModeSchema = z.enum(['light', 'dark', 'auto'])
+const resolvedThemeSchema = z.enum(['light', 'dark'])
+const themeKey = 'theme'
+
+type ThemeMode = z.infer<typeof themeModeSchema>
+type ResolvedTheme = z.infer<typeof resolvedThemeSchema>
+
+function getStoredThemeMode(): ThemeMode {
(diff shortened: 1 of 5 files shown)


src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { z } from 'zod'
+
+const themeModeSchema = z.enum(['light', 'dark', 'auto'])
+const resolvedThemeSchema = z.enum(['light', 'dark'])
+const themeKey = 'theme'
+
+type ThemeMode = z.infer<typeof themeModeSchema>
+type ResolvedTheme = z.infer<typeof resolvedThemeSchema>
+
+function getStoredThemeMode(): ThemeMode {
```
schiller-manuel: we could make this an isomorphic function using createIsomorphicFn

schiller-manuel: (this applies to the other functions as well)

src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+  return () => mediaQuery.removeEventListener('change', handler)
+}
+
+const themeDetectorScript = (function () {
+  function themeFn() {
+    try {
+      const storedTheme = localStorage.getItem('theme') || 'auto'
+
+      if (storedTheme === 'auto') {
+        const autoTheme = window.matchMedia('(prefers-color-scheme: dark)')
```
schiller-manuel, line 75: do we have a subscription to theme change on the client? e.g. if i switch from dark to light mode on the client in my browser or OS settings, does this update the theme here as well?

src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+          .matches
+          ? 'dark'
+          : 'light'
+        document.documentElement.classList.add(autoTheme, 'auto')
+      } else {
+        document.documentElement.classList.add(storedTheme)
+      }
+    } catch (e) {
+      const autoTheme = window.matchMedia('(prefers-color-scheme: dark)')
```
schiller-manuel, line 84: why do we call this again? when does the above throw?

src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+const ThemeContext = createContext<ThemeContextProps | undefined>(undefined)
+
+type ThemeProviderProps = {
+  children: ReactNode
+}
+export function ThemeProvider({ children }: ThemeProviderProps) {
+  const [themeMode, setThemeMode] = useState<ThemeMode>(getStoredThemeMode)
+
+  useEffect(() => {
```
schiller-manuel, line 108: do we really need that effect? cant we just execute it whenever setTheme is called?

Balastrong: I can move `updateThemeClass` inside `setTheme` but I'd keep the useEffect to un/subscribe to the system listener

src/components/ThemeProvider.tsx
Balastrong: Yes, it's the `setupPreferredListener` function setting it up. While on `auto` if you change the theme for example on your OS setting it immediately updates on tanstack.com

src/components/ThemeProvider.tsx
Balastrong: What is the advantage?

src/components/ThemeProvider.tsx
Balastrong: This is just in case accessing localStorage fails

PR #457, opened 2025-08-07
fix: raw links

Balastrong:
All the "get started" links are plain strings, not really typesafe in a tanstack style. This PR addresses it.

Also relative navigation does not infers the `libraryId` on pages where the param is defined, this might be material for an issue on Router though.

---

There are 101 typescript errors and 56 eslint warnings in the repo, I might take on this crusade step by step and get rid of them.

### DIFF
--- a/src/routes/_libraries/route.tsx
+++ b/src/routes/_libraries/route.tsx
@@ -275,7 +275,7 @@ function LibrariesLayout() {
         <div className="flex items-center gap-3 px-2 py-1 rounded-lg">
           <UserButton />
           <Link
-            to="/account"
+            to="/account/$"
             className="flex-1 text-sm font-medium text-gray-700 dark:text-gray-300 hover:text-gray-900 dark:hover:text-white"
           >
             My Account
--- a/src/routes/_libraries/start.$version.index.tsx
+++ b/src/routes/_libraries/start.$version.index.tsx
@@ -84,13 +84,21 @@ export default function VersionIndex() {
         </p>
         <div className="flex justify-center gap-4 flex-wrap">
           <Link
-            to="./docs/framework/react/quick-start#impatient"
+            from={'/$libraryId/$version'}
+            to={'./docs/framework/$framework/$'}
+            params={{
+              libraryId: library.id,
+              framework: 'react',
+              _splat: 'quick-start#impatient',
+            }}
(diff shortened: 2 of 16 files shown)


src/routes/_libraries/route.tsx
```diff
@@ -275,7 +275,7 @@ function LibrariesLayout() {
         <div className="flex items-center gap-3 px-2 py-1 rounded-lg">
           <UserButton />
           <Link
-            to="/account"
+            to="/account/$"
```
SeanCassiere, line 278: I'm unable to test this as I'm not on the list.

src/routes/_libraries/start.$version.index.tsx
```diff
@@ -84,13 +84,21 @@ export default function VersionIndex() {
         </p>
         <div className="flex justify-center gap-4 flex-wrap">
           <Link
-            to="./docs/framework/react/quick-start#impatient"
+            from={'/$libraryId/$version'}
+            to={'./docs/framework/$framework/$'}
+            params={{
+              libraryId: library.id,
+              framework: 'react',
+              _splat: 'quick-start#impatient',
```
SeanCassiere: Instead of adding heading id into the splat using `#`, it'd probably better if the `hash` prop is used instead.

```
hash="impatient"
```

src/routes/_libraries/route.tsx
SeanCassiere: It should still work fine though, since I think `/$` routes are still optional at the moment.

Balastrong: Typescript was flagging an error as the route `/account` isn't valid. The route file is at `routes/_library/account.$.tsx`.

Should the final `/$` be optional? In that case there's probably something to look at in the Router types.

src/routes/_libraries/start.$version.index.tsx
Balastrong: It is definitely better and more correct, thanks :)

src/routes/_libraries/route.tsx
SeanCassiere: Oh if that's the case, then yeah, what you have should be fine.

SeanCassiere: I'd just confirm the runtime behaves the same/as expected, then this can be resolved.

Balastrong: Oki, I tested and runtime also works. It's still an empty page for now anyway :)

PR #462, opened 2025-08-11
fix: enhance img component className handling

dborysov:


SeanCassiere (commented):
Was there anywhere where this specifically affected the content on the website?

SeanCassiere (approved):
Thanks!

PR #465, opened 2025-08-13
fix: counters in the homepage no longer disappear on first load

Balastrong:
We still have the counters in the homepage disappearing for a second right after the first load. No errors are thrown but the component still suspends and our fallback is an empty component.

Preloading the data will prevent the component from suspending right after mount.

While waiting for the current behaviour to be fixed upstream (see https://github.com/erquhart/convex-oss-stats/issues/7 and https://github.com/get-convex/convex-react-query/issues/22) this could be a nice temporary fix


erquhart (approved):
lgtm

PR #477, opened 2025-08-25
fix: normalize markdown links to work on both GitHub and website

KyleAMathews:
## Summary
- Adds path normalization to the `MarkdownLink` component to handle the difference between GitHub's file-based paths and the website's route-based paths
- Converts `./path` style links to `../path` to account for the website's URL structure where docs are at `/library/version/docs/file` (without trailing slash)
- This allows documentation authors to use GitHub-compatible relative links that also work correctly on tanstack.com

## Context
This fixes the issue described in https://github.com/TanStack/db/pull/449 where relative links between documentation pages require different paths for GitHub vs the website.

The problem occurs because:
- In GitHub: A link from `docs/overview.md` to `docs/guides/live-queries.md` uses `./guides/live-queries.md`
- On tanstack.com: The URL is `/library/version/docs/overview` (no trailing slash), so relative paths resolve differently and need `../guides/live-queries.md`

## Test plan
- [ ] Test that existing documentation links still work correctly on the website
- [ ] Verify that GitHub-style relative links (./path) now work on the website
- [ ] Check that absolute paths and external links are unaffected

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/components/MarkdownLink.tsx
+++ b/src/components/MarkdownLink.tsx
@@ -16,6 +16,14 @@ export function MarkdownLink({
 
   const [hrefWithoutHash, hash] = hrefProp?.split('#') ?? []
   let [to] = hrefWithoutHash?.split('.md') ?? []
+  
+  // Normalize relative paths that work in GitHub to work with our routing structure
+  // In GitHub: ./guides/foo.md works from docs/overview.md
+  // In our router: we need ../guides/foo because the current path is /lib/version/docs/overview (not overview/)
+  if (to?.startsWith('./')) {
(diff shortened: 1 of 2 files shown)


thruflo:
Cool!

Bit nervous matching slightly arbitrarily on `./`. Is there another way of somehow just starting the relative path traversal at the right level? Then other paths like `[#](foo.md)` and `[#](../foo.md)` can work too.

Also, whilst on this, the other relative links in GitHub are the file links like ../packages/db/src/foo.ts — I wonder if there’s a sensible way of detecting relative source file links and adding the right tree/blob GitHub URL prefix? Maybe too complex to be worth it.

PR #481, opened 2025-09-04
feat: add support for mermaid diagrams

SeanCassiere:

PR #485, opened 2025-09-17
migrate to @convex-dev/better-auth 0.8

erquhart:
Migrates Convex + Better Auth to latest, cleaned things up a bit. Will follow up with auth for SSR.


SeanCassiere (changes requested):
Could you run the lint script please?

erquhart:
@SeanCassiere I can but the failures are pre-existing. Example: https://github.com/TanStack/tanstack.com/pull/481

erquhart:
Pushed a fix:
- ignores dist and .content-collections so it doesn't fail erroneously on local
- rewrite a misformatted markdown file w/ prettier
- reformat two code files

Once this is merged your other PR's that are failing lint should succeed.

erquhart:
Added src/utils/gh-sponsor-meta.json to .prettierignore - don't know if this is the solution you want, but there's another one-off file in there too.

SeanCassiere (approved):

PR #495, opened 2025-09-25
feat: enhance library card hover interactions and fix badge visibility

fyzanshaik:
## Enhanced Library Card Hover Effects

Improves the homepage library cards with better hover animations and fixes badge visibility.

### Changes
- **Enhanced hover**: Cards lift with colored glow effects matching each library's theme
- **Fixed badge visibility**: Badges (RC, ALPHA, BETA) now stay visible above hover overlay
- **Smooth animations**: Added lift effect (`-translate-y-2`) and themed shadows (`shadow-current/20`)

### Problem Solved
Previously, badges would disappear behind the hover overlay, making it hard to see library status when hovering. Now badges stay visible with `z-40` instead of `z-20`.

### Result
More polished, interactive cards with better visual feedback and consistent badge visibility.

Screenshot From 2025-09-25 22-09-33

### DIFF
--- a/src/routes/_libraries/index.tsx
+++ b/src/routes/_libraries/index.tsx
@@ -141,9 +141,9 @@ function Index() {
               </p>
             </div>
           </div>
-        <div className="w-fit mx-auto px-4">
-          <OpenSourceStats />
-        </div>
+          <div className="w-fit mx-auto px-4">
+            <OpenSourceStats />
+          </div>
         </div>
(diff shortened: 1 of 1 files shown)


SeanCassiere (changes requested):
Would it be possible to fix this border issue on the cards which are hovered?

Image

Also, could you make the translate a bit more subtle?

fyzanshaik:
@SeanCassiere on it!

SeanCassiere (approved):

PR #497, opened 2025-09-26
fix: add scroll-padding to accommodate for the sticky header height

DamianOsipiuk:
Adds `scroll-padding-top` to accommodate for sticky header.

Otherwise all anchor links are scrolled too far and user would not see the header that was linked to as it's behind the header.

https://github.com/user-attachments/assets/bacb976e-1edc-4252-822b-92cb26c2074b


SeanCassiere (approved):
Confirmed this works locally, as our Convex preview builds are currently bugged out.

PR #508, opened 2025-10-10
finish up tanstack start rc upgrade

brenelz:
Biggest thing to test is using the netlify vite plugin. We could also use nitro v2 if we want to stay similar to previous
- Run formatter
- Setup and configure netlify plugin
- Update router/start packages to latest
- update vite to version 7
- Include relative link fix
- add convex defaults

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,7 +1,7 @@
 [context.production]
 command = "npx convex deploy --cmd 'pnpm run build'"
 [context.deploy-preview]
-command = "npx convex deploy --cmd 'pnpm run build' --preview-run"
+command = "npx convex deploy --cmd 'pnpm run build'"
 [build]
 command = "vite build"
 publish = "dist/client"


src/server/auth.server.ts
```diff
@@ -8,9 +8,7 @@ import { betterAuthComponent } from '../../convex/auth'
 // You'll want to replace this with an environment variable
 const siteUrl = process.env.URL
 const convexSiteUrl =
-  process.env.NODE_ENV === 'production'
-    ? process.env.VITE_CONVEX_SITE_URL!
-    : 'http://upbeat-greyhound-631.convex.site'
+  process.env.VITE_CONVEX_SITE_URL || 'http://upbeat-greyhound-631.convex.site'
```
schiller-manuel, line 11: we do have the default value already in the env schema, so can be removed here?

PR #524, opened 2025-10-31
improve auth stability, support ssr auth, login redirects

erquhart:
- Adds support and patterns for SSR compatible auth
- Adds redirect support

See /account page changes for patterns, /builder page changes for redirects

Issues I'd like some input on:
- We need a way to continue showing Query data from ssr until authentication state resolves for select queries.
- There's a flash of white when the user first authenticates, presumably due to a hydration mismatch, but I haven't found a clear fix for it

### DIFF
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import { redirect, createRouter } from '@tanstack/react-router'
 import { routerWithQueryClient } from '@tanstack/react-router-with-query'
 import { convexQuery, ConvexQueryClient } from '@convex-dev/react-query'
-import { ConvexProvider, ConvexReactClient } from 'convex/react'
+import { ConvexProvider } from 'convex/react'
 import { routeTree } from './routeTree.gen'
 import { DefaultCatchBoundary } from './components/DefaultCatchBoundary'
 import { NotFound } from './components/NotFound'
@@ -13,8 +13,7 @@ import { api } from 'convex/_generated/api'
 
 export function getRouter() {
   const CONVEX_URL = env.VITE_CONVEX_URL!
-  const convexClient = new ConvexReactClient(CONVEX_URL, { expectAuth: true })
-  const convexQueryClient = new ConvexQueryClient(convexClient)
+  const convexQueryClient = new ConvexQueryClient(CONVEX_URL)
 
   const queryClient: QueryClient = new QueryClient({
     defaultOptions: {


package.json
```diff
@@ -89,8 +89,7 @@
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "cmdk": "^1.1.1",
-    "convex": "^1.25.4",
-    "convex-oss-stats": "link:../../../erquhart/convex-oss-stats",
```
jherr, line 93: I don't think this is going to work.

erquhart: This updates the convex package and removes a non-existent file based package, was left in accidentally from a while ago. The actual package is @erquhart/convex-oss-stats, that dependency is still in place.

src/routes/builder.tsx
```diff
@@ -20,16 +21,19 @@ export const Route = createFileRoute('/builder')({
       description: 'Build amazing applications with TanStack',
     }),
   }),
+  beforeLoad: async (ctx) => {
+    requireAuth(ctx)
```
erquhart, line 25: Utility that checks for a user id, and if none is present, throws a redirect with the current location in the `redirectTo` query param.

src/routes/_libraries/account.tsx
```diff
@@ -7,22 +7,30 @@ import {
...
   component: AccountPage,
+  beforeLoad: async (ctx) => {
+    requireAuth(ctx)
+  },
+  loader: async (ctx) => {
+    await ctx.context.queryClient.ensureQueryData(
+      convexQuery(api.auth.getCurrentUser, {})
+    )
+    await ctx.context.queryClient.ensureQueryData(
+      convexQuery(api.llmKeys.listMyLLMKeysForDisplay, {})
```
erquhart, line 31: Need to parallelize these.

src/routes/__root.tsx
```diff
@@ -26,13 +26,32 @@ import { ToastProvider } from '~/components/ToastProvider'
...
+  const { createAuth } = await import('../../convex/auth')
+  const cookieNames = getCookieNames(createAuth)
+  const cookieAuth = getAuthFromCookie(getCookie(cookieNames.convexJwt))
+  if (cookieAuth) {
+    return cookieAuth
+  }
+  if (getCookie(cookieNames.sessionToken)) {
+    return await fetchAuth(getRequest())
+  }
+})
```
erquhart, line 54: Uses Convex jwt from cookie if it exists and is not expired, otherwise fetches a new one, which relies on the longer lived session cookie. If that fails the user is considered unauth.

src/routes/_libraries/account.tsx
```diff
@@ -532,24 +543,7 @@ function UserSettings() {
...
-          <p className="text-sm text-gray-600 dark:text-gray-300 mb-6">
-            Please sign in to access your account settings.
-          </p>
-          <Link to="/login">
-            <button className="text-sm font-medium bg-black/80 hover:bg-black text-white dark:text-black dark:bg-white/95 dark:hover:bg-white  py-2 px-4 rounded-md transition-colors">
-              Sign In
-            </button>
-          </Link>
-        </div>
-      </Unauthenticated>
```
erquhart, line 552: Dropped this as the page won't show for unauthenticated users.

PR #534, opened 2025-11-17
Create tanstack-router-route-matching-tree-rewrite.md

Sheraff:
New article related to this PR: https://github.com/TanStack/router/pull/5722

### DIFF
--- a/src/blog/tanstack-router-route-matching-tree-rewrite.md
+++ b/src/blog/tanstack-router-route-matching-tree-rewrite.md
@@ -31,7 +31,7 @@ root
     └── $id => match /users/$id
 ```
 
-We can add more routes to get a complete picture:
+Adding more routes gives a more complete picture:
 
 ```
 /users/$id
@@ -153,7 +153,7 @@ When building the segment trie, we need to parse each route (e.g., `/users/$user
 
 Instead of re-creating a new object every time, we can reuse the same object across all parsing operations to avoid allocations in the hot path.
 
-```tsx
+```ts
 const data = { kind: 0, prefixEnd: 0, suffixStart: 0, nextCursor: 0 }
 do {
   parseSegment(path, data)


src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
+---
+title: Performant Route Matching in TanStack Router
+published: 2025-11-18
+authors:
+  - Florian Pellet
+---
+
+![Big performance number](/blog-assets/tanstack-router-route-matching-tree-rewrite/header.png)
+
+We achieved a 20,000× performance improvement in route matching in TanStack Router. Let's be honest, this is *definitely* cherry-picked, but it's a real number using a real production application. And it demonstrates something important: matching a pathname to a route is no longer bottlenecked by the number of routes in your application.
```
LeCarbonator:
```suggestion
We achieved a 20,000× performance improvement in route matching in TanStack Router. Let's be honest, this is *definitely* cherry-picked, but the number is real and comes from a real production application. More importantly, it shows that matching a pathname to a route is no longer bottlenecked by the number of routes in your application.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+## The Real Problem: correctness, not speed
+
+One big responsibility of a router is to match a given URL pathname (e.g., `/users/123`) to a route definition (e.g., `/users/$userId`). This is deceptively complex when you consider all the different types of route segments (static, dynamic, optional, wildcard) and the priority rules that govern which route should match first.
+
+We used to have a route matching algorithm that would look through every route in the route tree, and through a mix of pattern matching, manual look-aheads, and recursion, find the best match. With the introduction of more features like optional segments and wildcards, the algorithm became increasingly complex and slow, and we started getting reports of incorrect matches.
```
LeCarbonator:
```suggestion
Our previous route matching algorithm would look through every route in the route tree, and through a mix of pattern matching, manual look-aheads, and recursion, find the best match. As we added more features like optional segments and wildcards, the algorithm became increasingly complex and slow, and we started receiving reports of incorrect matches.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+We opted for a complete rewrite: we now parse the route tree into a segment trie, and matching is done by traversing this trie. This makes it much simpler to implement exact matching rules while ensuring high performance.
```
LeCarbonator: Typo? I'm unfamiliar with trie, ignore this if it's an actual term.
```suggestion
We opted for a complete rewrite: we now parse the route tree into a segment tree, and matching is done by traversing this tree. This makes it much simpler to implement exact matching rules while ensuring high performance.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Algorithmic Complexity
+
+The reason we can get such a massive performance boost is because we've changed which variable drives the complexity of the algorithm. The bigger the route tree, the bigger the performance gain.
+
+- Old approach: `O(N)` where `N` is the number of routes in the tree.
+- New approach: `O(M)` where `M` is the number of segments in the pathname.
+
+Using this new trie structure, each check eliminates a large number of possible routes, allowing us to quickly zero in on the correct match.
```
LeCarbonator: Same as above, ignore if *trie* is a term.
```suggestion
Using this new tree structure, each check eliminates a large number of possible routes, allowing us to quickly zero in on the correct match.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+Say for example we have a route tree with 450 routes (pretty big app) and the tree can only eliminate 50% of routes at each segment check (this is unusually bad, it's often much higher). In 9 checks we have found a match (`2**9 > 450`). By contrast, the old approach *could* have found the match on the first check, but in the worst case it would have had to check all 450 routes, which yields an average of 225 checks. Even in this bad, simplified, and very unusual case, we are looking at a 25× performance improvement.
```
LeCarbonator:
```suggestion
For example, imagine we have a route tree with 450 routes (fairly large app) and the tree can only eliminate 50% of routes at each segment check (this is unusually bad, it's often much higher). With this bad setup, we have found a match in 9 checks (`2**9 > 450`). By contrast, the old approach *could* have found the match on the first check, but in the worst case it would have had to check all 450 routes, which yields an average of 225 checks. Even in this simplified case, we are looking at a 25× performance improvement.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+This is what makes tree structures so powerful.
+
+In practice, we've observed:
+- Small apps (10 routes): 60× faster
+- Big apps (450 routes): 10,000× faster
+
+These are lower than 20,000×, but they are still insane numbers. Yet they still aren't the full story.
```
LeCarbonator:
```suggestion
These are lower than 20,000×, but they are still insane numbers, and they're only part of the story.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Fun Implementation Details
+
+Beyond choosing the right data structures, working on performance is usually about avoiding death by a thousand cuts: avoiding memory allocations, skipping work, ... Here are some of the fun implementation details that helped us get to these numbers.
```
LeCarbonator: Use an ellipses or change the sentence.

Ellipses:
```suggestion
Beyond choosing the right data structures, working on performance is usually about avoiding death by a thousand cuts: avoiding memory allocations, skipping work, … Here are some of the fun implementation details that helped us get to these numbers.
```

Rephrased
```suggestion
Beyond choosing the right data structures, working on performance is usually about avoiding death by a thousand cuts: avoiding memory allocations, skipping work, and so on. Here are some of the fun implementation details that helped us get to these numbers.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Backwards Stack Processing
+
+We use a stack to manage our traversal of the trie, because the presence of dynamic segments (`/$required`, `/{-$optional}`, `/$` wildcards) means we may have multiple possible paths to explore at each segment.
```
LeCarbonator: See *trie* comments from earlier for reference.
```suggestion
We use a stack to manage our traversal of the tree, because the presence of dynamic segments (`/$required`, `/{-$optional}`, `/$` wildcards) means we may have multiple possible paths to explore at each segment.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+The ideal algorithm would be depth-first search (DFS) in order of highest priority, so that we can return as soon as we find a match. In practice, we have very few possibilities of early exit; but a fully static path should still be able to return immediately.
+
+To accomplish this, we use an array as the stack. But we know that pushing to and popping from the end of an array is O(1), while shifting and unshifting from the start is O(N). So we iterate candidates in reverse order of priority at each segment, pushing them onto the stack. This way, when we pop from the stack, we get the highest priority candidate first.
```
LeCarbonator: This needed a double take to read. I assume you mean that you avoid `shift`/ `unshift` because they're O(N)?

```suggestion
To accomplish this, we use an array as the stack. Since pushing and popping at the end of an array are O(1) operations, while shifting from the start is O(N), we avoid the latter entirely. At each segment, we iterate candidates in *reverse* order of priority, pushing them onto the stack. This way, when we pop from the stack, we get the highest priority candidate first.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+		}
+	}
+
+	// ... static segments last
+}
+```
+
+### Bitmasking for Optional Segments
+
+Optional segments introduce additional complexity, as they can be present or absent in the URL. While walking the trie, we need to track which optional segments were skipped (i.e. an array of booleans).
```
LeCarbonator: See trie comments above.

```suggestion
Optional segments introduce additional complexity, as they can be present or absent in the URL. While walking the tree, we need to track which optional segments were skipped (i.e. an array of booleans).
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+Every time we push onto the stack, we need to store the "state at which to pick up from" including which optional segments were skipped. But we don't want to have to `[...copy]` an array of booleans every time we push onto the stack as it would imply many short-lived allocations.
+
+To efficiently handle this, we use bitmasking to represent the presence or absence of each optional segment.
```
LeCarbonator:
```suggestion
To avoid this overhead, we use bitmasking to represent skipped optional segments.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+```ts
+const next = skipped | (1 << depth) // mark segment at 'depth' as skipped
+```
+
+And to read from the bitmask:
+```ts
+if (skipped & (1 << depth)) // segment at 'depth' was skipped
+```
+
+The downside is that this limits us to 32 segments, beyond what optional segments will never be considered skipped. We could extend this to a `BigInt` if needed, but for now, it's feels reasonable.
```
LeCarbonator:
```suggestion
The downside is that this limits us to 32 segments. Optional segments beyond that point will never be considered skipped. We could switch to a `BigInt` if needed, but for now, this feels reasonable.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+### Reusing Typed Arrays for Segment Parsing
+
+When building the segment trie, we need to parse each route (e.g., `/users/$userId/{-$maybe}`) into its constituent segments (e.g. `static:user`, `dynamic:userId`, `optional:maybe`). Doing this is basically running the same parsing algorithms hundreds of times, every time extracting the same structured data (i.e. segment type, value, prefix, suffix, where the next segment starts, etc).
```
LeCarbonator: Trie counter, see previous comments

```suggestion
When building the segment tree, we need to parse each route (e.g., `/users/$userId/{-$maybe}`) into its constituent segments (e.g. `static:user`, `dynamic:userId`, `optional:maybe`). Doing this is basically running the same parsing algorithms hundreds of times, every time extracting the same structured data (i.e. segment type, value, prefix, suffix, where the next segment starts, etc).
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+	if (cached) return cached
+	const result = performMatch(pathname)
+	cache.set(pathname, result)
+	return result
+}
+```
+
+With a cache, we only need to do the expensive matching operation once per unique pathname. Subsequent requests for the same pathname will be served from the cache, which is O(1).
+
+However, as we've seen before, some apps can have a very large number of unique routes, which means even more unique pathnames (e.g., route `/user/$id` is matched by `/user/1`, `/user/2`, etc). To prevent unbounded memory growth, we implement a Least Recently Used (LRU) cache that automatically evicts the least recently used entries when the cache reaches a certain size.
```
LeCarbonator:
```suggestion
However, as we've seen before, some apps can have a very large number of unique routes, which means even more unique pathnames (e.g., route `/user/$id` is matched by `/user/1`, `/user/2`, etc). To prevent unbounded memory growth, we implement a Least Recently Used (LRU) cache. When the cache reaches a certain size, it automatically evicts the least recently used entries.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+![buildLocation performance over 4 evolutions of the algorithm](/blog-assets/tanstack-router-route-matching-tree-rewrite/buildlocation-evolution-benchmark.png)
+
+Even the smallest apps see some improvement here, but it might not feel as dramatic. We will continue to optimize the other parts of the router to make it feel as snappy as we can. But at least route matching is no longer a bottleneck!
+
+## Going even further
+
+While we are very happy with these results (and are probably done optimizing route matching for now), there are still some avenues we could explore to push this even further:
+
+- **sub-segment nodes**: currently, each node in the trie represents a full URL segment (between slashes). We could further break down segments into sub-segments (e.g., prefix, dynamic part, suffix) to allow for much better branch elimination. This is what [`find-my-way`](https://github.com/delvedor/find-my-way) does (the router behind [Fastify](https://www.fastify.io/)), and it yields impressive results.
```
LeCarbonator, line 224: Trie counter, see previous comments
```suggestion
- **sub-segment nodes**: currently, each node in the tree represents a full URL segment (between slashes). We could further break down segments into sub-segments (e.g., prefix, dynamic part, suffix) to allow for much better branch elimination. This is what [`find-my-way`](https://github.com/delvedor/find-my-way) does (the router behind [Fastify](https://www.fastify.io/)), and it yields impressive results.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+![benchmark results for LRU cache vs Object vs Map](/blog-assets/tanstack-router-route-matching-tree-rewrite/lru-benchmark.png)
+
+## The full story
+
+The numbers we've presented so far are impressive. They're also cherry-picked from the biggest apps we tested, which is biased in favor of the new algorithm. They're also comparisons against the old, uncached algorithm. In reality, we've added caching a while ago. We can see the full progression over the last 4 months:
+
+![route matching performance over 4 evolutions of the algorithm](/blog-assets/tanstack-router-route-matching-tree-rewrite/matching-evolution-benchmark.png)
+
+And besides that, they also focus on a small part of the router's performance profile. Matching a pathname to a route is only one part of the job. If we look at a "fuller" functionality, for example `buildLocation`, which involves matching, building the location object, interpolating the path, passing the validation functions, running the middlewares, etc, we see a more modest but still significant improvement:
```
LeCarbonator:
```suggestion
And besides that, they also focus on a small part of the router's performance profile. Matching a pathname to a route is only one part of the job. If we look at a more "complete" operation, for example `buildLocation`, which involves matching, building the location object, interpolating the path, passing the validation functions, running the middlewares, etc, we see a more modest but still significant improvement:
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+
+![buildLocation performance over 4 evolutions of the algorithm](/blog-assets/tanstack-router-route-matching-tree-rewrite/buildlocation-evolution-benchmark.png)
+
+Even the smallest apps see some improvement here, but it might not feel as dramatic. We will continue to optimize the other parts of the router to make it feel as snappy as we can. But at least route matching is no longer a bottleneck!
```
LeCarbonator:
```suggestion
Even the smallest apps see some improvement here, but it might not feel as dramatic. We will continue to optimize the other parts of the router to make it feel as snappy as we can. The good news is route matching is no longer a bottleneck!
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,184 @@
...
+While we are very happy with these results (and are probably done optimizing route matching for now), there are still some avenues we could explore to push this even further:
+
+- **sub-segment nodes**: currently, each node in the trie represents a full URL segment (between slashes). We could further break down segments into sub-segments (e.g., prefix, dynamic part, suffix) to allow for much better branch elimination. This is what [`find-my-way`](https://github.com/delvedor/find-my-way) does (the router behind [Fastify](https://www.fastify.io/)), and it yields impressive results.
+
+- **branch compression**: we could also expand in the other direction, and have tree nodes that represent multiple segments at once, when there is no branching (usually, for a series of static segments). This would reduce the depth of the tree and the number of stack frames we need to process.
+
+---
+
+This wasn't a "let's make it faster" project, it was a "let's make it correct" project that happened to yield massive performance improvements as a side effect. I rarely have the opportunity to see such big numbers when I benchmark stuff, so I hope you didn't mind my cherry-picking a bit for this post.
```
LeCarbonator:
```suggestion
This wasn't a "let's make it faster" project, it was a "let's make it correct" project that happened to yield massive performance improvements as a side effect. I rarely see numbers this large in real benchmarks, so I hope you’ll forgive a bit of cherry-picking in this post.
```

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,230 @@
...
+
+![Big performance number](/blog-assets/tanstack-router-route-matching-tree-rewrite/header.png)
+
+We achieved a 20,000× performance improvement in route matching in TanStack Router. Let's be honest, this is _definitely_ cherry-picked, but the number is real and comes from a real production application. More importantly, it shows that matching a pathname to a route is no longer bottlenecked by the number of routes in your application.
+
+## The Real Problem: correctness, not speed
+
+One big responsibility of a router is to match a given URL pathname (e.g., `/users/123`) to a route definition (e.g., `/users/$userId`). This is deceptively complex when you consider all the different types of route segments (static, dynamic, optional, wildcard) and the priority rules that govern which route should match first.
+
+Our previous route matching algorithm would look through every route in the route tree, and through a mix of pattern matching, manual look-aheads, and recursion, find the best match. As we added more features like optional segments and wildcards, the algorithm became increasingly complex and slow, and we started receiving reports of incorrect matches.
```
schiller-manuel: can we bring this in:

base of the previous algorithm was trying to sort all routes into a flat list.
since our sorting did not adhere to a strict weak ordering, sorting was not well defined and even behaved differently between Chrome and Firefox

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,230 @@
...
+
+[See implementation.](https://github.com/TanStack/router/blob/f830dffb7403819ea984017bb919b4a8708f24a5/packages/router-core/src/lru-cache.ts)
+
+This data structure performs about half as well as a regular `Object` for writes, and on par with an `Object` for reads. This is a trade-off we are willing to make to avoid unbounded memory growth.
+
+![benchmark results for LRU cache vs Object vs Map](/blog-assets/tanstack-router-route-matching-tree-rewrite/lru-benchmark.png)
+
+## The full story
+
+The numbers we've presented so far are impressive. They're also cherry-picked from the biggest apps we tested, which is biased in favor of the new algorithm. And they're comparisons against the old, uncached algorithm. In reality, we've added caching a while ago. We can see the full progression over the last 4 months:
```
schiller-manuel: "They're" => "They are" (2x)

schiller-manuel: same for "we've" => "we have"

src/blog/tanstack-router-route-matching-tree-rewrite.md
```diff
@@ -0,0 +1,230 @@
...
+
+While we are very happy with these results (and are probably done optimizing route matching for now), there are still some avenues we could explore to push this even further:
+
+- **sub-segment nodes**: currently, each node in the trie represents a full URL segment (between slashes). We could further break down segments into sub-segments (e.g., prefix, dynamic part, suffix) to allow for much better branch elimination. This is what [`find-my-way`](https://github.com/delvedor/find-my-way) does (the router behind [Fastify](https://www.fastify.io/)), and it yields impressive results.
+
+- **branch compression**: we could also expand in the other direction, and have tree nodes that represent multiple segments at once, when there is no branching (usually, for a series of static segments). This would reduce the depth of the tree and the number of stack frames we need to process.
+
+---
+
+This wasn't a "let's make it faster" project, it was a "let's make it correct" project that happened to yield massive performance improvements as a side effect. I rarely see numbers this large in real benchmarks, so I hope you’ll forgive a bit of cherry-picking in this post.
```
schiller-manuel: this is the first time "I" is used. probably better to stay with "We"

PR #538, opened 2025-11-20
Update query.gg banners for Black Friday sale

lynnandtonic:
To merge **November 23rd**

Note: I was getting an error for the Query landing page, so could you check the rendering of the banner there? Nothing changed from the previous time, but just in case. Thanks! 💚


lynnandtonic:
@LadyBluenotes Never mind, I know what's going on with the CSS and can push a change. I'll submit a new PR.

PR #603, opened 2025-12-25
fix: ensure reactStartHandler executes client-side only to avoid window access error

sahilkhedkar:
The `TypeError: Cannot read properties of undefined (reading 'window') occurs when @convex-dev/better-auth's reactStartHandler` is executed in a server-side context during development. The error originates from the `Request `constructor (likely undici in Node.js) attempting to access `window`, which is undefined on the server. This happens because the handler contains client-side code that assumes a browser environment.

The issue manifests when the handler is bundled in dependencies and called server-side (e.g., in a TanStack Start server function or loader), but works when inlined and executed client-side (e.g., in a React component's useEffect).

Solution
Ensure reactStartHandler runs only on the client by calling it in a client-side context, such as:

Inside a React component with `useEffect` to defer execution until after mount.
In a client component (if using client/server boundaries).
Avoid importing it in server functions or loaders to prevent server-side execution.
This prevents the handler's client-dependent code from running in the Node.js environment of the TanStack Start dev server.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,0 +1,3391 @@
+{
+  "lockfileVersion": 1,
+  "workspaces": {
+    "": {
+      "name": "tanstack.com",
+      "dependencies": {
+        "@auth/core": "0.37.0",
+        "@floating-ui/react": "^0.27.8",
+        "@headlessui/react": "1.7.18",
+        "@netlify/functions": "^5.1.0",
+        "@netlify/neon": "^0.1.0",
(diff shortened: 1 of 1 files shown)


SeanCassiere:
This doesn't seem to have anything to do with the tanstack.com marketing and documentation website.

Based on the content in the description, I think you may want to first head over to the `#router-questions` channel in the [TanStack Discord server](https://tlinz.com/discord) and ask your question there. Alternatively, you could head over to the [TanStack/Router](https://github.com/tanstack/router) repo and post an issue there.

PR #615, opened 2025-12-30
Fix: remove padding on the right sidebar so it doesn't get cut off

lukamircetic:
### Problem:
The right sidebar with the partner info was being cut off in my browser:
image

### Fix:
- removed the left padding so that it wouldn't be overflowing

image

### Testing:
- [x] checked that this fix works on multiple window sizes including mobile
- [x] checked that this fix works in multiple browsers (chrome, safari, firefox)


lukamircetic:
Looks fixed in main now, closing.

PR #631, opened 2026-01-07
fix: replace blog ISO dates with readable dates

LeCarbonator:
A wrongly formatted string caused the date formatter to fall back to ISO dates:

image

This PR changes the formatting strings to be `MMM d, yyyy` instead of `MMM dd, yyyy` for a readable date string.

(Image from local environment)

image


dandaka:
@LeCarbonator please don't kill my first PR to OSS 🥹

PR #669, opened 2026-01-17
Modernize README to be more aligned with other TanStack READMEs

Floriferous:
Started contributing to this repo, and wanted to add a small note on the need to also clone router/query/table for the husky hooks to pass, and then felt like making the README a bit closer to the other tanstack repos.

Obviously a bit of a bigger change, not sure if this is welcome, but happy to adjust or let someone more experienced take over from here and get the details right :)

A few ideas:
* Might be nice to add a banner at the top like the other repos, but I don't have the right template to do that I believe.
* Since people would only come to this repo to edit the docs, and not use a library, I thought that adding a contributing.md is an unnecessary layer of indirection, but it's a bit less standard compared to the other tanstack repos. Focus is on quickly getting setup and able to submit a PR

Can be previewed here: https://github.com/Floriferous/tanstack.com/tree/docs/update-readme

## Summary by CodeRabbit

* **Documentation**
  * Restructured README with a cleaner landing page layout
  * Simplified local development setup instructions
  * Added TanStack Ecosystem exploration section with community resources
  * Enhanced "Get Involved" guidance for contributors


LadyBluenotes (approved):
Great change! Could you update the conflicts and then we can get this merged?

PR #676, opened 2026-01-21
Fix Feed page for search predicates with overlay array predicates

Julienng:
Hi!

When using the page https://tanstack.com/feed?page=1&pageSize=50&viewMode=table&libraries=%5B%22router%22%5D

The query in Drizzle is failing because of an incorrect query generation:

```sql
select "id", "entry_id", "entry_type", "title", "content", "excerpt", "published_at", "created_at", "updated_at", "metadata", "library_ids", "partner_ids", "tags", "show_in_feed", "featured", "auto_synced", "last_synced_at"
from "feed_entries"
where (
       "feed_entries"."show_in_feed" = $1
  and  "feed_entries"."published_at" >= $2
  and  "feed_entries"."published_at" <= $3
  and  "feed_entries"."library_ids" && ARRAY[$4]::text[]
)--                                    ~~~~~~~~~
--                                         ^-- sql driver is not happy with that
order by "feed_entries"."published_at" DESC
-- params: true,1970-01-01T00:00:00.000Z,2026-01-21T18:35:09.396Z,router
```

By looking and searching at the Drizzle doc, I've found `arrayOverlaps(posts.tags, ['Typescript', 'ORM'])` (https://orm.drizzle.team/docs/operators#arrayoverlaps)

When looking at the generated query, it looks like this:

```sql
select "id", "entry_id", "entry_type", "title", "content", "excerpt", "published_at", "created_at", "updated_at", "metadata", "library_ids", "partner_ids", "tags", "show_in_feed", "featured", "auto_synced", "last_synced_at"
from "feed_entries"
where (
       "feed_entries"."show_in_feed" = $1
  and  "feed_entries"."published_at" >= $2
  and  "feed_entries"."published_at" <= $3
  and  "feed_entries"."library_ids" && $4
)
order by "feed_entries"."published_at" DESC
-- params: true,"1970-01-01T00:00:00.000Z","2026-01-21T18:45:32.506Z","{\"start\"}"
```

So the query's initial behavior is maintained but without an SQL error.

## PR Status:
- I have no data locally to test the query behavior; it should be the same (&& predicate is kept)
- The SQL error is fixed
- I was blocked because of the pre-commit hook locally; I had to `--no-verify` to get to the PR. *need a check if everything is good*
- I've migrated all `ARRAY[${sql.join` patterns, but two of them needed a cast for the enum values: `filters.useCases` and `data.capabilityFilter`

### DIFF
--- a/src/utils/feed.server.ts
+++ b/src/utils/feed.server.ts
@@ -111,9 +111,6 @@ export function buildFeedQueryConditions(
     filters.libraries.length > 0
   ) {
     conditions.push(arrayOverlaps(feedEntries.libraryIds, filters.libraries))
-    // conditions.push(
-    //   sql`${feedEntries.libraryIds} && ${filters.libraries}::text[]`,
-    // )
   }
 
   // Partner filter (array overlap) - use parameterized array


SeanCassiere (approved):
👏🏼 thank you!

PR #720, opened 2026-02-17
fix: Transform Solid docs hook calls correctly

GEREGUR:
## Summary
- Adds `solidTransform.ts` with regex-based transformation to convert React-style `useQuery({...})` to Solid-style `useQuery(() => ({...}))`
- Integrates transformation in `documents.server.ts` before the broken string replacements  
- Skips the incorrect string replacement rules for Solid hooks since the transformation handles it properly

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,11 +115,17 @@
     "zustand": "^4.5.2"
   },
   "devDependencies": {
+    "@babel/generator": "^7.29.1",
+    "@babel/parser": "^7.29.0",
+    "@babel/traverse": "^7.29.0",
+    "@babel/types": "^7.29.0",
     "@content-collections/core": "^0.8.2",
     "@content-collections/vite": "^0.2.4",
     "@eslint/js": "^9.39.1",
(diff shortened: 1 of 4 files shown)


SeanCassiere (changes requested):
I think this is something, where the actual docs for the Project needs to be changed instead.

This change, in this PR, would be applied to all projects on all markdown files - making it impossible for a comparison style doc or to just circumvent this transform if needed.

We keep the our markdown parsing to the bare minimum and only add features which make it easier for the maintainers of the individual TanStack project write out their docs.

---

I'd suggest, opening a PR to Query or the docs of the project's you think needs correction. We also, have some special frontmatter we use for text-replacement allowing less duplicated content - [see here](https://github.com/TanStack/query/blob/bb257c2835ba12475ac083b10763216280c58e3a/docs/framework/solid/guides/dependent-queries.md?plain=1#L5-L13).

TkDodo:
closing because stale

PR #818, opened 2026-04-09
Add aria-label to copy page dropdown trigger

MoonBrillante:
## Summary

Adds an `aria-label` to the icon-only dropdown trigger next to the "Copy page" button so screen readers announce it more clearly.

## Changes

- add an `aria-label` to the dropdown trigger button
- derive the `aria-label` from the `label` prop so it stays consistent when the label is overridden

## Notes

I only changed this one component. I wasn't able to run the full local lint successfully because my local setup hit unrelated repository-wide issues, including a missing `content-collections` module and a Node version mismatch.

Related to #793

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accessibility of the copy-page dropdown trigger by adding a descriptive aria-label, enhancing compatibility with assistive technologies and screen readers.


LadyBluenotes:
Thank you for doing this! Do you mind updating the branch then we can get this merged?

MoonBrillante:
Done, I’ve updated the branch. Thanks!

LadyBluenotes (approved):

PR #851, opened 2026-04-22
fix(docs): keep notFound inside route boundaries

Varkoff:
## Summary

Opening the official docs page for structured outputs currently breaks:

- Docs URL: [https://tanstack.com/ai/latest/docs/guides/structured-outputs](https://tanstack.com/ai/latest/docs/guides/structured-outputs)
- Runtime error: `Unexpectedly client reference export 'notFound' is called on server`

The underlying problem is that the shared docs helpers and `createServerFn` handlers were throwing `notFound()` directly. Once that crossed the server function boundary, SSR blew up before the route loader could do its normal redirect or 404 handling.

This patch keeps `notFound()` inside route boundaries:

- shared docs helpers and docs server functions now throw a small docs-specific not-found error
- route loaders treat that error the same way they already treat route-level not-found cases
- the markdown route handlers translate that docs-specific error back into `notFound()` at the route layer

That keeps the redirect and 404 behavior in the route layer, and it stops the server function layer from leaking router `notFound()` across the boundary.

## Screenshot

Structured outputs docs error

## Test plan

- [x] `pnpm run test:tsc`
- [x] `pnpm run test:lint`
- [x] `pnpm run test:smoke`

## Summary by CodeRabbit

* **Refactor**
  * Unified missing-doc detection so all doc handlers recognize the same “docs not found” condition.
* **Bug Fixes**
  * Doc requests that fail this check now try to resolve a redirect (308) and otherwise return a not-found response.
* **New Features**
  * Added a dedicated docs-not-found error type and guard to standardize handling across routes and helpers.


schiller-manuel:
notFound works again as expected now

PR #893, opened 2026-05-06
fix(og): externalize @takumi-rs/core so Netlify ships the native binary

AlemTuzlak:
## Summary

`https://tanstack.com/api/og/<library>.png` returns **200 OK with `content-type: image/png` and a zero-byte body** in production — the request succeeds but the cached PNG is empty, so Slack/Twitter/etc. show a broken image. Locally everything works.

### Root cause

`@takumi-rs/core` uses napi-rs's runtime-dispatched native binding loader (`createRequire(import.meta.url)` + platform-conditional `require()` of `@takumi-rs/core-<platform>-<arch>-<libc>/*.node`). Netlify's zip-it-and-ship-it bundles the SSR function with esbuild, which can't statically trace those optional platform binaries — so the Linux `.node` file is missing from the lambda bundle.

At runtime the binding load fails. The error is thrown inside `ImageResponse`'s `ReadableStream.start(controller)` callback, but at that point the `Response(stream, ...)` was already constructed with status 200 and `image/png`. `controller.error()` errors the body stream, so Netlify's runtime emits an empty body — and Netlify Edge happily caches the 200 + 0 bytes for `s-maxage=86400`. Hence the symptom.

```
HTTP/2 200
content-type: image/png
content-length: 0
cache-status: "Netlify Edge"; hit
```

### Fix

- **`netlify.toml`** — add `external_node_modules = ["@takumi-rs/core"]` so Netlify ships `node_modules/@takumi-rs/core` (and the matching `@takumi-rs/core-linux-x64-gnu` that pnpm installs on the Linux build machine via optional deps) as-is. napi-rs's runtime dispatcher then resolves the binary normally instead of bundling failing.
- **`src/routes/api/og/$library[.png].ts`** — `await result.ready` before returning, so any future render failure surfaces as a real `500` instead of a silently-cached empty 200. (Without this, even a transient renderer error would poison the edge cache.)

## Test plan

- [x] Local smoke tests: `pnpm test:smoke` — both OG endpoints render real PNGs (`OG image · library landing (284748 bytes)`, `OG image · docs page (273744 bytes)`)
- [x] `pnpm test` (tsc + lint) clean on changed files
- [ ] After merge: `curl -sI https://tanstack.com/api/og/ai.png` returns `content-length > 0` and the image renders in og:image previewers (Slack unfurl, Twitter Card validator)
- [ ] Confirm Netlify deploy logs do not show "Failed to load @takumi-rs/core in Node.js runtime"

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Open Graph image generation by awaiting readiness, adding error logging, and returning proper 500 responses when rendering fails.
  * Enhanced Open Graph image URL resolution to reliably derive origin from live server-side requests for SSR contexts.

* **Chores**
  * Updated deployment configuration to ensure platform-specific native binaries are packaged as external dependencies.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,20 +7,18 @@ publish = "dist/client"
 
 [functions]
 directory = "netlify/functions"
+# OG image rendering goes through @takumi-rs/wasm (forced via the `module`
+# option in src/server/og/generate.server.ts) instead of @takumi-rs/core's
+# native napi binding — Netlify's function bundler dropped the platform-
+# specific .node optional dep no matter how we configured it, and WASM
+# sidesteps the whole binary-resolution dance. The .wasm asset isn't part
+# of the JS import graph, so include it explicitly.
 included_files = [
   "public/fonts/Inter-Regular.ttf",
   "public/fonts/Inter-ExtraBold.ttf",
   "public/images/logos/splash-dark.png",
+  "node_modules/.pnpm/@takumi-rs+wasm@*/node_modules/@takumi-rs/wasm/pkg/takumi_wasm_bg.wasm",
 ]
-# @takumi-rs/core ships platform-specific .node binaries via napi-rs's runtime
-# require() dispatcher — esbuild can't statically trace the optional deps. Keep
-# the napi loader external so node_modules/@takumi-rs/core ships as-is, and
-# also list the Linux x64 binary package explicitly: Netlify's bundler only
-# follows declared deps for `external_node_modules`, so the optional platform
-# package would otherwise be dropped (verified — function failed at runtime
-# with "Cannot find native binding"). Netlify functions run on AWS Lambda
-# Amazon Linux 2 (glibc, x64), hence linux-x64-gnu.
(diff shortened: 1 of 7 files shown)


LadyBluenotes (approved):

kane50613:
@AlemTuzlak hey I'm the maintainer of Takumi, didn't saw it getting merged and used.

hope my original #625 proposal can help fix the [netlify.toml](https://github.com/kane50613/tanstack.com/blob/main/netlify.toml) for napi package and [.npmrc](https://github.com/kane50613/tanstack.com/blob/main/.npmrc) to ensure it installed properly and works.

I just released the 2.0 beta. Some things might change, but the setup should still be very similar.

Thanks for reading this

AlemTuzlak:
Hey @kane50613, I wasn't actually aware that there was a PR to do this already, I just really liked takumi and wanted to generate custom og images, if you'd like to go over the current implementation on main and see if anything can be improved that would be awesome!

PR #907, opened 2026-05-09
fix(shop/Select): move 'onKeyDown' from wrapper 'div' to trigger 'button' for a11y

sukvvon:
## Summary

- The wrapper `<div>` in `ShopSelect` had an `onKeyDown` handler without a `role`, which violates the `jsx-a11y/no-static-element-interactions` rule and causes lint to fail
- Moves the handler to the trigger `<button>`, which is always the focus holder in this component (focus stays on the trigger; option `<button>` hover only updates a `focused` state, not DOM focus)
- Preserves the keyboard navigation (arrows, Enter, Escape) introduced in #898 — no user-facing behavior change
- Resolves the lint failure that has been blocking CI on PRs branched from main

## Test plan

- [ ] Open `/shop` and use the sort dropdown via keyboard: Tab to focus, Enter/Space/ArrowDown to open, Arrow keys to move, Enter to select, Escape to close
- [ ] Confirm `pnpm test` passes locally with `Found 10 warnings and 0 errors.`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard navigation behavior for the select component.


sukvvon:
I think this issue already fixed. so close this PR.

PR #917, opened 2026-05-12
draft

LadyBluenotes:


src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+draft: false
+excerpt: "A companion to our incident postmortem: what we're changing across the org so the May 11 supply-chain attack can't happen the same way again."
+authors:
+  - Sarah Gerrard
+  - Corbin Crutchley
+  - Jack Herrington
+  - Tanner Linsley
+---
+
+This week, fourteen of our packages were quietly republished to npm with malware baked in. None of us authored those releases. None of us approved them. By the time the first report landed in our issue tracker, the malicious versions had already been sitting on the registry for a while.
```
Sheraff: What is "for a while"? (20 minutes or 2 days?)

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+
+We've already published [the full incident postmortem](/blog/npm-supply-chain-compromise-postmortem), but if you want the timeline, the attack chain, the exact package list, the IOCs, and the "what to do if you installed a bad version" guidance, that's the source of truth. Read that first.
```
Sheraff: The `but` feels weird here
```suggestion
We've already published [the full incident postmortem](/blog/npm-supply-chain-compromise-postmortem), and if you want the timeline, the attack chain, the exact package list, the IOCs, and the "what to do if you installed a bad version" guidance, that's the source of truth. Read that first.
```

src/blog/incident-followup.md
schiller-manuel: not quite correct?. we initiated those releases by merging to main and thus causing our release pipeline to run.

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+
+That's the safeguard stack. It's the same kind of stack that, on a normal day, makes "steal a maintainer's npm token and publish malware" a non-starter, and most npm supply-chain attacks you've read about in the last two years are some variation of that exact scenario, which is why those defences exist and why we'd invested in them.
+
+This incident happened because the attacker didn't try to defeat any of those safeguards directly. They didn't steal a token, bypass 2FA, forge a signature or sneak code past review. They engineered a path where our pipeline would steal its own token, for them, at the exact moment it was minted, by way of a cache that everyone in the chain implicitly trusted. That's a sophisticated piece of work, and pretending otherwise wouldn't be honest either.
```
crutchcorn: Let's expand this section more. Specifically, using the OIDC process prevents:

-A single-point-of-failure NPM publish token (which didn't happen and is why this incident 'only' hit our Router packages)

- The kind of malware spreading that's occurring through local machine secret stealing

- Making it harder for us to scope who can release and when

- Making audit logs harder to trace down (which was a huge help with us figuring out this issue)

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+
+We've already published [the full incident postmortem](/blog/npm-supply-chain-compromise-postmortem), but if you want the timeline, the attack chain, the exact package list, the IOCs, and the "what to do if you installed a bad version" guidance, that's the source of truth. Read that first.
+
+This post is the companion piece. The postmortem covered what happened. This one is about what we're changing because of it.
+
+## The shortest possible recap
+
+Just enough context so this post makes sense on its own:
+
+Someone opened a pull request from a throwaway fork. The PR looked unremarkable, but it triggered a workflow that checked out the contributor's code and ran it in a job that had write access to our shared CI cache. The contributor's code poisoned the cache. Later, an entirely legitimate merge to main ran our release workflow, which restored the poisoned cache, extracted our short-lived publish token out of the runner's memory, and used it to push 84 malicious versions across our router-family packages.
```
schiller-manuel: "The PR looked unremarkable,"

we never really got a chance to even know this PR exists at all, it was immediately closed and thus not visible by default in githubs PR view

schiller-manuel: wouldnt use "contributor" here. "attacker" is more like it

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+- **Upgraded every repo to pnpm 11**, so we inherit the ecosystem install-cooldown behaviour. It's not a fix on its own. It buys us a window.
+
+Most of those are blast-radius reductions rather than root-cause fixes; they shrink what a similar attack could reach, but the workflow shape itself is what we're rebuilding in the next round.
+
+## What we're working on next
+
+Some of the changes we're making take more than a hotfix to land properly, and a few of them are still open questions we're working through together. We're including the unresolved ones because how we're thinking about them feels as relevant as what we've already shipped.
+
+- **Adding [`zizmor`](https://github.com/woodruffw/zizmor) as a required PR check on every repo.** It's a static analyser for GitHub Actions workflows, and if we'd had it running against the workflow this incident exploited, it almost certainly would have flagged the `pull_request_target` + fork-checkout pattern before any of this happened.
```
Sheraff: Would it have flagged it? It would be nice to know instead of "almost certainly"

src/blog/incident-followup.md
schiller-manuel: "and used it to push 84 malicious versions across our router-family packages." 

it was two indivual CI runs that were used to publish 42 releases each

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+- **Putting `CODEOWNERS` on `.github` folders**, restricting workflow changes to core maintainers. We're leaning toward "yes" on this one; workflow files are code that runs with our credentials, and that means they should be treated like the most privileged code in the repo, because that's effectively what they are.
+- - **Replacing the pnpm setup cache with `actions/cache/restore`**, which has more conservative defaults: explicit restore, no implicit write-on-exit. The auto-save behaviour of `actions/cache@v5` is what allowed the cache write to happen regardless of our `permissions:` block, and we want that path closed by design instead of by accident.
```
Sheraff:
```suggestion
- **Replacing the pnpm setup cache with `actions/cache/restore`**, which has more conservative defaults: explicit restore, no implicit write-on-exit. The auto-save behaviour of `actions/cache@v5` is what allowed the cache write to happen regardless of our `permissions:` block, and we want that path closed by design instead of by accident.
```

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+
+That's the safeguard stack. It's the same kind of stack that, on a normal day, makes "steal a maintainer's npm token and publish malware" a non-starter, and most npm supply-chain attacks you've read about in the last two years are some variation of that exact scenario, which is why those defences exist and why we'd invested in them.
+
+This incident happened because the attacker didn't try to defeat any of those safeguards directly. They didn't steal a token, bypass 2FA, forge a signature or sneak code past review. They engineered a path where our pipeline would steal its own token, for them, at the exact moment it was minted, by way of a cache that everyone in the chain implicitly trusted. That's a sophisticated piece of work, and pretending otherwise wouldn't be honest either.
+
+## The honest part
```
schiller-manuel, line 48: please change this AI jargon ...

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+
+The workflow pattern that made this possible — [`pull_request_target`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) plus a checkout of fork code plus a cache write — has been [documented as a known-bad pattern by GitHub's own security team](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) for over three years. We had it in production. We knew about the pattern in the abstract, and never connected it to our own workflows. That's a thing we have to sit with.
+
+There's a version of this post where we lean on the fact that npm provenance, SLSA, OIDC, and 2FA all worked exactly as advertised and still didn't stop this. That's true, and it matters.
```
schiller-manuel: i dont get this sentence

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+
+However, modern supply-chain defences are not enough on their own, and anyone telling you otherwise is selling something. But it's also a deflection. The hole was in our workflow design. We're the ones who left it open. So that's where the changes are going.
+
+## What we've already done
+
+These are the changes that have landed since the incident. They're not the whole plan — they're the things we could do quickly.
+
+- **Removed the cache from our pnpm setup**, temporarily. Caching is coming back, but not in the form that got us into this.
```
schiller-manuel: "temporarily", "coming back".
are we sure? i am not

src/blog/incident-followup.md
```diff
@@ -0,0 +1,87 @@
...
+- **Putting `CODEOWNERS` on `.github` folders**, restricting workflow changes to core maintainers. We're leaning toward "yes" on this one; workflow files are code that runs with our credentials, and that means they should be treated like the most privileged code in the repo, because that's effectively what they are.
+- - **Replacing the pnpm setup cache with `actions/cache/restore`**, which has more conservative defaults: explicit restore, no implicit write-on-exit. The auto-save behaviour of `actions/cache@v5` is what allowed the cache write to happen regardless of our `permissions:` block, and we want that path closed by design instead of by accident.
+- **Isolating the cache between release and PR environments.** Even with the safer restore action in place, we don't want a PR-context job to ever write into a cache namespace that a release-context job will read. Different keys, different scopes, different worlds.
```
schiller-manuel: or we just dont cache release at all?

Sheraff (commented):
I think this is good and straight to the point. 

If i were a hater I would say that the "what we're working on next" doesn't feel like enough: it patches *this* type of security issues, but what are we doing for our *broader* security posture?

src/blog/incident-followup.md
tannerlinsley: 20 minutes is correct I believe and sounds better than "a while"

src/blog/incident-followup.md
crutchcorn: Yeah, I think we should be more reflective of how we're thinking about this. There's an active unknown if cache needs to return or not but it needs more research. Regardless of where that research takes us, we have it disabled for the time being.

src/blog/incident-followup.md
tannerlinsley:
```suggestion
This week, fourteen of our packages were quietly republished to npm with malware baked in. None of us authored those releases. None of us approved them. By the time the first report landed in our issue tracker, the malicious versions had already been sitting on the registry for ~20 minutes.
```

crutchcorn: Yeah, also I wouldn't personally categorize how long it was there as `a while` - AFAIK it was a very short window

20jasper (commented):
I think tonally everything is good and responsibility is well accepted. I think this has a bit too much overlap with the postmortem article linked up front. Consider a quick "this was a cache poisoning and permissions attack caused by this notorious GHA feature, read more on the root cause here"

PR #973, opened 2026-06-09
fix(routes/examples): remove fixed 'h-[95dvh]' breaking Safari render

sukvvon:
## Summary

Removes the fixed `h-[95dvh]` height from the examples page scroll container, which prevented the embedded code example viewer from rendering in Safari.

## Problem

On example pages, the interactive code explorer rendered correctly in Chrome but collapsed in Safari. The scroll container used a fixed `h-[95dvh]` inside a `md:flex-row` flex layout. WebKit miscalculated this fixed `dvh` height, collapsing the container to ~79px while its content was ~1005px tall. The `overflow-auto` then clipped everything below the collapsed height.

## Fix

```diff
-    <div className="flex-1 flex flex-col min-h-0 overflow-auto h-[95dvh]">
+    <div className="flex-1 flex flex-col min-h-0 overflow-auto">
```

The container already receives its height from the parent flex layout via `flex-1 min-h-0`, so `h-[95dvh]` was redundant in Chrome (height unchanged) and harmful in Safari. Verified the fix renders correctly in both Safari and Chrome.

#### Safari

image

#### Chrome

image

#### Safari

image

#### Chrome

image

## Summary by CodeRabbit

* **Style**
  * Adjusted the layout of the examples section to provide better vertical flexibility by removing a fixed height constraint while maintaining responsive scrolling behavior.

### DIFF
--- a/src/routes/$libraryId/$version.docs.framework.$framework.examples.$.tsx
+++ b/src/routes/$libraryId/$version.docs.framework.$framework.examples.$.tsx
@@ -351,7 +351,7 @@ function PageComponent() {
   }
 
   return (
-    <div className="flex-1 flex flex-col min-h-0 overflow-auto h-[95dvh]">
+    <div className="flex-1 flex flex-col min-h-0 overflow-auto">
       <div className="p-4 lg:p-6">
         <DocTitle>
           <span>


bytaesu:
👍

PR #984, opened 2026-06-15
fix(npm-stats): stop summary cards overcounting co-installed packages

tombeckenham:
## Problem

The NPM Stats summary cards (All-time, Monthly, Weekly, Daily) were inflated by ~9×. For TanStack Start, Weekly showed **148.1M** when the real figure is **~16.5M**.

**Cause:** the cards summed *every* `@tanstack/*start*` package mapped to the library (43 of them). ~9 of those — `start-client-core`, `start-server-core`, `start-plugin-core`, `react-start`, etc. — are transitive dependencies of a single `@tanstack/react-start` install, so each install was counted ~9 times. The chart directly below the cards was already correct because it uses the library's declared `npmPackageNames`.

## Fix

Make the summary cards honor `npmPackageNames` too, so they match the chart:

- **Daily/Weekly/Monthly** — `NPMSummary` now uses the shared `recentDownloadsQuery` (which passes `npmPackageNames`) instead of a local copy that dropped it and fell back to the full registered-package set.
- **All-time** — `rebuildOssStatsCache` now restricts each library's aggregate to its declared `npmPackageNames` when present (libraries without an explicit list are unchanged).

## Testing

No DB required — the stats code falls back to live npm data when there's no cache. Run `pnpm dev`, open `/start/latest/docs/npm-stats`: Weekly now shows ~16.5M (matching the chart) instead of 148.1M.

> Note: the all-time card reads a precomputed cache row in production, so it corrects on the next `rebuildOssStatsCache` run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Bug Fixes**
  * NPM statistics (downloads and package counts) are now calculated more accurately by only counting explicitly included packages for each library, excluding transitive dependencies that were previously inflating totals.


KevinVandy (approved):

sukvvon (approved):

KevinVandy (dismissed):
If we change the weekly number, wouldn't we have to change the monthly and daily numbers?

KevinVandy:
Let's have @tannerlinsley review the download number changes before this gets merged. I think I agree with the change, but this will be a quite visible change in how we've been displaying the download numbers.

Issue #1069, opened 2026-07-29
Form API link in header = 404


bvinnerd:
Clicking on Form's API link takes the user to a 404 page.

To reproduce:
- Go to https://tanstack.com/form/latest
- Click **API**

The page displays a 404 error.

tannerlinsley:
Verified against production on July 31, 2026: the exact Form API link (/form/latest/docs/reference/index) now returns 200, as do the related root and framework reference routes. The runtime fix shipped on main in 1e23c84b, so this report is resolved.

PR #1072, opened 2026-07-29
Test docs index-file fallback

Mohith26:
## What changed

`main` already shipped the runtime fix in `1e23c84b`: `loadDocs` retries a missing `<path>.md` as `<path>/index.md`.

After resolving the conflict, this PR now keeps that implementation and adds focused regression coverage for:

- canonical root and framework reference paths loading their index files
- ordinary docs loading with one request and no fallback
- non-404 fallback failures propagating

The only production-code addition is a typed optional fetch function so the loader can be tested without changing its default runtime behavior.

## Evidence

This covers the root cause reported in [TanStack/pacer#237](https://github.com/TanStack/pacer/issues/237) and tracked in #1069. The affected live root and framework reference URLs now return 200 responses.

## Validation

- `pnpm test`
- 119 tests: 118 passed, 1 skipped
- TypeScript and type-aware lint clean

## Risk

Low. Runtime path resolution remains the implementation already on `main`; this PR primarily pins its behavior with tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved loading for index-style documentation pages (including framework-specific reference sections) by retrying with an `index.md` fallback when the direct document isn’t found.
  - Updated route redirect behavior to resolve redirects directly after a not-found result, without the previous extra canonicalization step.

- **Tests**
  - Reworked the test suite to validate the document-loading flow directly, including the requested path sequence, direct non-index loading, successful index fallback, and correct error propagation when fallback fails.


tannerlinsley:
Thanks for the work on this. The runtime fix has already shipped on main, and we’ve reverified that the affected Form and Pacer reference routes are returning 200. Since the remaining value here is regression coverage rather than an unshipped fix, we’re going to close this PR and keep the review queue focused. Appreciate the contribution.

PR #1074, opened 2026-07-29
feat(home): add a calm and pleasant hero loop

wiesson:
# feat(home): add a calm and pleasant hero loop

## What

The homepage hero palm now breathes. Same shot, same framing, but as a 12 second
video loop with a very gentle push-in — roughly 3% scale over the full cycle, so
it reads as "alive" rather than "animated".

Because why not.

## How it loops

The source is a one-directional zoom, so `loop` on its own would visibly snap
back to frame 0. The shipped clip is a ping-pong instead — forward `0 → 144`,
then reverse `143 → 1` — which makes the wrap seamless in both directions.

Measured as SSIM between the last and first frame:

| | last → first frame |
|---|---|
| raw source | 0.914 |
| ping-pong | 0.985 |
| adjacent frames mid-clip (baseline) | 0.990 |

## Poster and fallback

`poster` is frame 0 of the loop itself, not the existing
`hero-palm-gradient-1600.webp`. The two crops differ slightly, and using the old
still would cause a visible jump the moment playback starts.

If no `<source>` can be decoded, the `<video>` just keeps showing the poster —
which is exactly the current homepage. There is no worse-than-today state.

## Encoding

The sky is a single continuous gradient, which bands badly under 8-bit
compression. Both files are therefore 10-bit:

| File | Codec | Size | Serves |
|---|---|---|---|
| `hero-palm-gradient-loop.webm` | VP9 profile 2, 10-bit, crf 38 | 959 KB | Chrome, Firefox, Edge |
| `hero-palm-gradient-loop.mp4` | HEVC Main10 (hvc1), crf 28 | 897 KB | Safari (HW decode on Apple silicon) |
| `hero-palm-gradient-loop.webp` | frame 0 | 24 KB | poster / fallback |

8-bit H.264 needed roughly 2.7 MB to reach comparable gradient smoothness, so
10-bit is both the better-looking and the smaller option here. Audio and the
cover-art stream are stripped.

The codec strings are written out explicitly in `type`:

```
video/webm; codecs="vp09.02.30.10"
video/mp4;  codecs="hvc1.2.4.L93.B0"
```

Without them Safari reports generic `video/webm` as playable, selects the WebM,
and then fails on the 10-bit VP9. With them, source selection resolves correctly
on the first try. Verified in Chromium via `canPlayType` and actual playback.

## Reduced motion

`prefers-reduced-motion: reduce` pauses the loop on its first frame, so those
users get the still hero. This mirrors how `PixelSpinner` already handles the
same preference. `autoplay` stays declarative so the loop does not wait on
hydration; the effect only walks it back.

`disableRemotePlayback` keeps Safari from offering an AirPlay overlay on what is
a decorative background.

## Assets kept

Nothing is deleted. `hero-palm-gradient-960.webp` is still used by
`src/routes/ds.stats.tsx` and `-1600.webp` by
`scripts/generate-brand-assets.mjs` for the OG image.

## Known limitation

The source clip is 1168×768, below the 1600/2400 ladder the site ships for the
still. `object-cover` crops about 25% of the height in the ~2:1 hero card and
scales the remainder by ~1.33×, which absorbs most of it, but on a high-density
display it is softer than the current photo. Happy to re-encode from a
higher-resolution source if that matters for merge.

## Summary by CodeRabbit

* **New Features**
  * Updated the home page hero area with looping video playback and a poster image.
  * Added support for multiple video formats for broader browser compatibility.

* **Accessibility**
  * Respects reduced-motion preferences by pausing the video and displaying its opening frame.


tannerlinsley:
Thanks for putting this together — the careful work on the loop, encoding, fallbacks, and reduced-motion behavior is appreciated. We’re going in a different direction for the homepage hero, and Andy is working on an entirely new animated hero image that will replace this approach, so I’m going to close this PR. Thanks again for the contribution!

PR #1094, opened 2026-08-02
refactor(*): migrate deprecated phosphor icon names to 'Icon' suffix

sukvvon:
`@phosphor-icons/react` v2.1.10 marks every unsuffixed export as `@deprecated` in favor of its `*Icon` counterpart (e.g. `Star` → `StarIcon`). Both names resolve to the same component, so this is a rename with no runtime behavior change — it clears the deprecation strikethrough in editors and stops new code from picking up the old names.

## Changes

- **180 files** — rewrote imports from `@phosphor-icons/react` and their usage sites to the `*Icon` exports. Aliased imports keep their local name (`Sparkle as Sparkles` → `SparkleIcon as Sparkles`).

- **`scripts/generate-phosphor-icon-registry.mjs`** — imports the suffixed export for every icon. This also drops the `Infinity` special case, which existed only because the bare name collides with the JS global; `InfinityIcon` never does.

- **`src/components/ds/phosphor-icons.generated.ts`** — regenerated from the updated script rather than hand-edited.

- **`src/routes/ds.iconography.lazy.tsx`** — the icon browser copied `<Acorn />` to the clipboard. The registry keys icons by their bare name, so the copy string and tooltip now append `Icon` to match what callers actually import. Search and the visible grid labels still use the bare name.

Inline `code={...}` examples on the `ds/*` pages were updated as well, so the JSX shown on the page matches the supported import.

## Notes

Icon names that collide with unrelated components (`Link`, `Table`, `Spinner`, `Command`, `User`) were left alone where they resolve to something other than Phosphor — only identifiers imported from `@phosphor-icons/react` were touched.


SeanCassiere (approved):
Looks like until they ship their next major version, we'll need to be vigilant about this.

Lucide has a nice convenience of being able to do this which made it so that typescript rejected imports without the icon suffix.

```ts
// lucide-react.d.ts
declare module 'lucide-react' {
  export * from 'lucide-react/dist/lucide-react.suffixed'
}
```

sukvvon:
@SeanCassiere Good call — I verified the approach works for Phosphor too.

Phosphor doesn't ship a suffixed-only entrypoint like Lucide's, so `export * from '...suffixed'` has nothing to point at: the barrel does `export * from './csr/Star'` ×1512, and each icon module exports both `Star` and `StarIcon`. But declaring the module with only the suffixed members gets the same result:

```
probe.ts(2,10): error TS2305: Module '"@phosphor-icons/react"' has no exported member 'Star'.
```

We already generate a registry from the same `dist/csr` scan, so emitting a `.d.ts` alongside it is straightforward. One catch: the ambient declaration replaces the barrel's types wholesale, so it has to re-export `Icon`, `IconProps`, `IconWeight`, `IconContext`, `IconBase` and `SSR` as well, or anything using those breaks.

Leaving it out of this PR so this one stays a pure rename.

PR #1123, opened 2026-08-06
refactor(*): migrate 'radix-ui' primitives to 'base-ui'

sukvvon:
Replaces all `@radix-ui` primitives with Base UI and removes the three `@radix-ui/react-{dialog,dropdown-menu,tooltip}` dependencies.

## Component mapping

| Radix | Base UI |
|---|---|
| `Dialog.Overlay` | `Dialog.Backdrop` |
| `Dialog.Content` | `Dialog.Popup` |
| `DropdownMenu.Content` | `Menu.Positioner` + `Menu.Popup` |
| `Tooltip.Content` | `Tooltip.Positioner` + `Tooltip.Popup` |
| `Tooltip.Provider` `delayDuration`/`skipDelayDuration` | `delay`/`timeout` |
| `DropdownMenu.Item` `onSelect` | `Menu.Item` `onClick` |
| `onSelect` + `e.preventDefault()` | `closeOnClick={false}` |
| `onInteractOutside` + `preventDefault()` | `onOpenChange` + `eventDetails.cancel()` (root-only) |
| `data-[state=open&#124;closed]` | keyframes: `data-open` / `data-closed`<br>transitions: `data-starting-style` / `data-ending-style` |
| `--radix-*-transform-origin` | `--transform-origin` |
| `--radix-dropdown-menu-content-available-height` | `--available-height` |
| `focus:` styling on items | `data-highlighted:` |

## `asChild` → `render`

The dropdown adapters drop Radix's `asChild` boolean in favour of Base UI's `render` element prop.

On triggers `render` is **required** and `children` is not in the type. That matters: 22 of 24 `<DropdownTrigger>` call sites passed no `asChild` prop at all and silently relied on its `= true` default, so a boolean rename would have flipped them to wrapper mode with no compile error. Making `render` required turns every one of those sites into a type error instead.

On items `render` stays optional and `children` is kept — their previous default was already falsy, so wrapper-mode call sites are unchanged.

`DropdownTrigger` also gains an optional `nativeButton` passthrough. Base UI defaults it to `true`, so a trigger whose `render` is not a native `<button>` would otherwise be given button semantics — Radix's `asChild` had no such notion. Two call sites need `nativeButton={false}`: `BrandContextMenu`, whose trigger is a `position: fixed` virtual anchor `<span>`, and `AuthenticatedUserMenu`, whose trigger is a `<div>` wrapping the avatar.

## Radix-only workarounds removed

`SearchModal` no longer needs the machinery that worked around Radix unmounting content during its exit animation — Base UI keeps popups mounted through it natively. Removed: `forceMount`, the `shouldRenderSearch` mount gate, `searchModalTransitionMs`, and the `document.body.style.pointerEvents` save/restore block.

That last one was not merely redundant: Radix portaled content outside the body's pointer-events scope, Base UI does not. Left in place it rendered the search modal visible but unclickable.

## Overlay primitives from `main`

`main` landed `Dialog`, `Drawer` and `Takeover` in `src/components/ds/ui/` after this branch opened, migrating nine overlays onto them. All three were Radix-backed, so they are ported here, along with the eleven `overlay-audit` specimens that copy the shells verbatim.

The port is the same mapping as everywhere else, plus one API removal. `onInteractOutside` was exposed on all three `*Content` components, but Base UI routes every dismissal through the root's `onOpenChange` — there is nowhere for a content-level prop to hook in. Only one call site passed it: `LibrariesOverlay`, keeping a portaled menu inside the takeover from closing it. That guard now lives in the `onOpenChange` it already had:

```tsx
if (
  details.reason === 'outside-press' &&
  details.event.target instanceof Element &&
  details.event.target.closest('[role="menu"]')
) {
  details.cancel()
}
```

`DialogContent` and `DrawerContent` had no consumers for the prop, so it is dropped rather than bridged to a root-level context. `/ds/takeover`'s notes and `specimen-meta`'s verdicts are updated to match.

`BuilderGuideDialogSpecimen` also loses its animation classes. They came from tailwindcss-animate, which is not installed, and were keyed off `data-state` — so they animated nothing before this branch and would animate nothing after it.

## Radix DOM contracts consumed outside the migrated files

Radix's `data-state` and `--radix-*` are a DOM contract, so anything reading them breaks when the primitive is swapped — with no compile error. Four places consumed them:

- `src/styles/app.css` animation selectors for `.cart-panel`, `.cart-overlay`, `.search-modal-overlay`, `.search-modal-content` and `.dropdown-content`
move from `[data-state='open'|'closed']` to `[data-open]` / `[data-closed]`. Base UI documents two idioms — `data-open`/`data-closed` for keyframe animations and `data-starting-style`/`data-ending-style` for transitions — and these rules are keyframe-based, so they use the former. `Tooltip` and the builder tooltips animate with transitions and use the latter.
- `PackagePills` sized its menu with `--radix-dropdown-menu-content-available-height`, which no longer resolves;
now `--available-height`. Confirmed in the browser that Base UI sets it on the positioner and it inherits to the popup.
- `LibraryLayout` keeps the docs sidebar open while a dropdown launched from inside it is active, via
`expandedMenuRef.current?.querySelector('[data-state="open"]')`. Base UI marks the open **trigger** with `data-popup-open`; the popup itself is portaled to `body` and so is out of that subtree, which rules out `[data-open]` here. Verified against the live sidebar: the new selector matches and the old one does not.
- `main`'s overlay keyframes — 20 selectors across `[data-ds-dialog-*]`, `[data-ds-drawer-*]`, `[data-ds-takeover-*]` and `[data-ds-pop]` — were keyed off `data-state` too. Left alone, every overlay animation the audit added would have silently stopped, with `tsc`, lint and the tests all still passing. The tooltip's three open states (`open`/`delayed-open`/`instant-open`) collapse to a single `data-open`, since Base UI does not distinguish the delay.

All but the first were found while merging `main`, not by `tsc` — which is the concrete form of the silent-failure risk this section describes.

One more landed the same way, caught in review: `PackagePills` kept four `onClick` handlers calling `e.preventDefault()`. Under Radix that cancelled the close; under Base UI `preventDefault()` on a plain click handler does nothing, so its group-visibility toggle, colour picker, sub-package toggles and *Add Packages* would each have dismissed the menu on first use. They now use `closeOnClick={false}` like `BaselineSection`. The same edit removed an `as unknown as React.MouseEvent` cast that the old signature had required.

## Other

- The npm-stats and charts menus use `Menu` from `@base-ui/react/menu` directly rather than through a wrapper. An earlier revision of this branch kept a `DropdownMenu*` adapter so those call sites needed no edits, but three of its four parts only renamed Base UI's API (`DropdownMenu` → `Menu.Root`, `DropdownMenuTrigger` → `Menu.Trigger`, `DropdownMenuItem` → `Menu.Item` with `onSelect` in place of `onClick`), which hid the real primitives for no benefit. `Portal`/`Positioner`/`Popup` is now written out at each of the 11 surfaces. The existing `Dropdown.tsx` is untouched — it carries the site's default popup styling, which these call sites do not want.
- `/ds/dropdown`, `/ds/cards` and `/ds/buttons` copy and code samples updated to reference Base UI.

## Verification

`tsc`, lint and the test suite pass.

Exercised in the browser, locally and on the preview deploy: blog author filter (open → select → filter applied), `/ds/dropdown`, `/ds/buttons` split button, navbar social menu, `CopyPageDropdown`, breadcrumb TOC, `BrandContextMenu` (anchors at cursor), `LoginModal`, `CartDrawer`, and `SearchModal` — including that its framework filter opens inside the modal without dismissing it. Console clean.

The preview deploy reaches the npm API, so the stats dropdowns were exercised there: all four `ChartControls` menus, `BaselineSection` presets, `LatestBucketNavigator` playback, and `NPMStatsChart` export/embed.

Things `tsc` cannot check were confirmed by reading computed style rather than by eye:

- The docs sidebar guard: opening the sidebar's own `FrameworkSelect` and evaluating both selectors against that subtree — `[data-popup-open]` matches, `[data-state="open"]` does not.
- `PackagePills`' `max-height` resolves to a real value again (`1719px`, matching `--available-height` on the positioner). With the stale `--radix-*` variable it computed to `none`.
- `CartDrawer` plays `cart-panel-in` on open, then takes `data-closed` and plays `cart-panel-out` / `cart-overlay-out` on close — the CSS attribute rename works in both directions. The search modal's four animations were confirmed the same way via `getAnimations()`.

`closeOnClick={false}` was checked where it replaces Radix's `onSelect` + `preventDefault()` idiom: clicking a baseline preset, and clicking the custom-ms input inside the playback menu, both leave the menu open.

Keyboard operation was checked directly, since `Menu.Item` takes `onClick` where Radix took `onSelect`: focusing an item in the blog author menu and pressing Enter, and separately Space, both apply the filter and close the menu. On `/ds/dropdown`, Down opens the menu and highlights the first item, Down moves the highlight, Enter selects and returns focus to the trigger, and Escape closes. Base UI handles all of that internally — no extra key handling was needed.

The overlay primitives were exercised on the preview deploy at `/ds/dialog`, `/ds/drawer`, `/ds/overlays` and `/ds/takeover`. Enter and exit animations both play (`dialog-panel-in` / `dialog-panel-out`, `drawer-right-in` at 280ms with `data-side="right"` matching), with focus trap, focus restoration, scroll lock and Escape confirmed on each. The `[role="menu"]` guard was checked end to end in both places it matters: selecting a framework in the libraries takeover, and a library in the search modal — the filter applies and the overlay stays open.

Read out of the deployed stylesheet rather than by eye: all 24 overlay animation rules resolve to `[data-open]` / `[data-closed]` with no `data-state` left, and the four `prefers-reduced-motion` blocks survive the rewrite.

The `maxHeight` menu still scrolls, and a text input nested inside the npm-stats package menu accepts typing without Base UI's typeahead stealing the keystrokes.

Not exercised: `AuthenticatedUserMenu` and `AvatarCropModal` (need auth) and `charts/ChartControls` (admin only). Each is the same code path as something above — `AvatarCropModal` differs from the verified `LoginModal` only in `max-w-md` vs `max-w-xs`, `AuthenticatedUserMenu` uses the same `DropdownItem render={...}` as the navbar social menu, and `charts/ChartControls` has the same `Menu.Positioner` line as the verified `npm-stats/ChartControls`. The one thing that inference does not cover is `AvatarCropModal`'s `react-easy-crop` widget interacting with Base UI's focus trap; the closest evidence is `SearchModal`'s nested filter menu working.

## Summary by CodeRabbit

* **Refactor**
  * Standardized menus, dialogs, tooltips, and dropdowns across the application.
  * Preserved existing navigation, selection, filtering, editing, chart controls, and modal workflows.
  * Improved consistency for popup positioning, transitions, and interactions across search, library, login, cart, and other overlays.
  * Updated dropdown interactions and visual states for more consistent behavior throughout the app.

* **Documentation**
  * Updated design-system descriptions and examples to reflect the refreshed menu and tooltip experiences.


LadyBluenotes:
Could you fix the dropdown styling, please? it's got rounded corners on some sides but not others 
image

sukvvon:
@LadyBluenotes Thanks! Measured this preview and production `main` — identical: `border-radius: 8px` on all four corners, all four actually round on my machine. Not something the Base UI migration introduced; `main` has the same `max-h-80 overflow-auto` on the rounded element.

Likely the scrollbar — I'm on macOS overlay scrollbars (`offsetWidth - clientWidth === 2`), so a space-consuming track would square off the right corners for you but stay invisible to me. Do you have "Show scroll bars: Always" on, or Windows/Linux? If so I'll move the scroll container inside the rounded popup.


https://refactor-radix-to-base-ui-tanstack-com.thetanstack.workers.dev/blog
image
image

PR #1139, opened 2026-08-11
Autoplay runnable docs examples when visible

tannerlinsley:
## What changed

- defer runnable documentation examples with TanStack Start hydration boundaries
- hydrate only near the viewport, then autoplay after visible dwell and browser idle time
- keep offscreen workbench and esbuild chunks cold
- cancel pending autoplay during focus, pointer, wheel, background-tab, or viewport-exit interaction
- preserve static SSR source and reserve the final workbench height

## Why

Runnable examples looked static until readers noticed and clicked Run. This makes visible examples come alive without adding the editor/runtime cost to initial page load.

## Validation

- `pnpm test` — 125 passed, 1 skipped
- production build passed
- production-mode browser smoke verified static-first loading, visible autoplay to Ready, offscreen deferral, and immediate manual Run


tannerlinsley:
Landed directly on main in 1d8d5b0f.

PR #1220, opened 2026-09-05
fix(builder): unlock BYOK passkey on the click, not mid-send

jherr:
Saved BYOK keys could stall Builder sends because the passkey unlock started too late in the send pipeline. Jack traced the failure to expired browser user activation.

## Fix

- Start unlock from the shared submission handler, for both composer messages and preview comments.
- Block duplicate submissions and composer/model edits while unlock is pending.
- Discard pending submissions after a conversation, credential scope, or model change. Restore preview comments when unlock fails.
- Only clear the composer when it still contains the submitted message.
- Separate local transcript hydration from synced-project updates. Local hydration previously retriggered itself and kept Send disabled.

## Testing

- `pnpm test`: passed, including type checks, lint, and 478 passing tests, with one skipped test.
- `node scripts/test-builder-passkey-submit.mjs`: passed against a local dev server. It uses an isolated browser and a fake encrypted key to check duplicate submission, disabled editing during unlock, and draft preservation after cancellation.
- Jack's original investigation covered fresh-click versus delayed unlock in Dia/Safari. The automated browser check mocks the passkey ceremony, it does not test 1Password itself.

## Companion SDK change

https://github.com/TanStack/ai/pull/1332 adds the failed-activation error for existing keyrings while preserving first-time registration's follow-up PRF ceremony.

## Attribution

Original fix and investigation by Jack Herrington. Follow-up fixes are added to his existing PR without replacing his commits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate submissions while passkey unlocking is in progress.
  * Ensured the submission continues with the provider selected when unlocking begins.
  * Disabled the composer, submit button, and model picker during passkey unlocking.
  * Preserved draft text when a passkey unlock is canceled.
  * Prevented outdated in-progress unlocks from affecting newer conversations, credential scopes, or model selections.

### DIFF
--- a/scripts/test-builder-passkey-submit.mjs
+++ b/scripts/test-builder-passkey-submit.mjs
@@ -1,0 +1,122 @@
+import assert from 'node:assert/strict'
+import { chromium } from 'playwright-core'
+
+// Run against the local dev server: node scripts/test-builder-passkey-submit.mjs
+const origin = process.env.BUILDER_TEST_ORIGIN ?? 'http://127.0.0.1:3008'
+assert.ok(['localhost', '127.0.0.1'].includes(new URL(origin).hostname))
+const browser = await chromium.launch({ channel: 'chrome', headless: true })
+try {
+  const page = await browser.newPage()
+  await page.addInitScript(() => {
+    class Passkey {
+      rawId = new Uint8Array([1]).buffer
+      getClientExtensionResults() {
+        return {
+          prf: { enabled: true, results: { first: new Uint8Array(32) } },
+        }
+      }
+    }
+    window.passkeyCalls = 0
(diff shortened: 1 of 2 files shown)


jherr:
Addressed both CodeRabbit findings in 541af78a:

- **Major (async unlock window):** added an `unlockingRef` in-flight guard so a repeat submit during the pending passkey ceremony is ignored, and read `provider` once before the await so a mid-await model change can't retarget the unlock.
- **Minor (silent abort):** on the still-locked path, set `error` ("Could not unlock the API key. Try again.") instead of returning silently.

Kept it to the re-entrancy guard rather than fully disabling the model picker during unlock, to keep the diff minimal. `test:tsc` and `lint:code` pass.

